PR #57, opened 2021-12-06
fix(redux): make `getBagItems` return empty array

gabrielfmp:
## Description

Our `getBagItems` selector is returning `undefined` when the bag or bag items don't exist but that
output breaks consistency and might create an error in other selectors that use `getBagItems` without
validating if it is an array or undefined. To validate if the bag is fetched or not we have other,
more specific, selectors.

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/bags/__tests__/selectors.test.ts
+++ b/packages/redux/src/bags/__tests__/selectors.test.ts
@@ -107,6 +107,31 @@ describe('bags redux selectors', () => {
 
       expect(selectors.getBagItems(mockState)).toEqual(expectedResult);
     });
+
+    it('should return an empty array when there are no bag or bag items', () => {
+      const mockStateWithoutBagItems = {
+        ...mockState,
+        bag: {
+          error: null,
+          id: null,
+          isLoading: false,
+          result: null,
+          items: {
+            ids: null,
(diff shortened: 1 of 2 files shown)


anapim (approved):

nelsonleite (approved):

ricardojmf (approved):

packages/redux/src/bags/__tests__/selectors.test.ts
```diff
@@ -201,14 +226,14 @@ describe('bags redux selectors', () => {
       expect(selectors.getBagItemsUnavailable(mockState)).toHaveLength(0);
     });
 
-    it('should return null if there is no bag loaded', () => {
+    it('should return empty array if there is no bag loaded', () => {
```
catarinacouto, line 229:
```suggestion
    it('should return an empty array if there is no bag loaded', () => {
```

catarinacouto: Here isn't better to expect a `.toEqual([])`? Just to be according to the test description?

catarinacouto (changes requested):

catarinacouto (approved):

PR #104, opened 2022-01-04
feat(client|redux): conversion(forms) js into ts

danielbento92:
## Description

- Conversion to typescript:
 - Client/forms 
 - Redux/forms

## Checklist

## Disclaimer


packages/client/src/forms/getFormSchema.ts
```diff
@@ -19,16 +10,25 @@ import join from 'proper-url-join';
...
  * instance (axios).
  *
  * @returns {Promise} Promise that will resolve then the call to the endpoint finishes.
+ *
  */
-export default (schemaCode, query, config) =>
-  client
+
+const getFormSchema: GetFormSchema = (
+  schemaCode: string,
```
boliveira, line 22: No need to type the parameters for the function as they are already typed by the `GetFormSchema` type. So, you can remove the typings from `schemaCode` and `query` parameters.

packages/client/src/forms/postFormData.ts
```diff
@@ -13,10 +14,16 @@ import client, { adaptError } from '../helpers/client';
  *
  * @returns {Promise} Promise that will resolve then the call to the endpoint finishes.
  */
-export default (schemaCode, data, config) =>
+const postFormData: PostFormSchema = (
+  schemaCode: string,
```
boliveira, line 18: Same as https://github.com/Farfetch/blackout/pull/104/files#r778044823

packages/client/src/forms/types/formSchema.types.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * @typedef {object} FormSchema
```
boliveira, line 2: No need for @typedefs when you are defining a typescript type with the same name. better-docs will generate the jsdoc from the typescript type automatically.

packages/client/src/forms/types/formSchemaQuery.types.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * @typedef {object} FormSchemaQuery
```
boliveira, line 2: Same as https://github.com/Farfetch/blackout/pull/104/files#r778058698

packages/redux/src/forms/actions/factories/fetchFormSchemaFactory.ts
```diff
@@ -34,9 +41,14 @@ import * as actionTypes from '../../actionTypes';
  *
  * @returns {FetchFormSchemaThunkFactory} Thunk factory.
  */
-export default getFormSchema =>
-  (schemaCode, query, config) =>
-  async dispatch => {
+const getFormSchema =
+  (getFormSchema: GetFormSchema) =>
+  (
+    schemaCode: string,
```
boliveira, line 47: Same as https://github.com/Farfetch/blackout/pull/104/files#r778044823

packages/redux/src/forms/actions/factories/fetchFormSchemaFactory.ts
```diff
@@ -34,9 +41,14 @@ import * as actionTypes from '../../actionTypes';
  *
  * @returns {FetchFormSchemaThunkFactory} Thunk factory.
  */
-export default getFormSchema =>
-  (schemaCode, query, config) =>
-  async dispatch => {
+const getFormSchema =
```
boliveira, line 44: This should be called `fetchFormSchemaFactory` and I would give it a type instead of typing the individual constituents. Also, make sure you reuse the same type for the parameters of the factory function as the parameters for the client function as they are the same in this case.

packages/redux/src/forms/actions/factories/submitFormDataFactory.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @function submitFormDataFactory
+ * @memberof module:forms/actions/factories
+ *
+ * @param {Function} postFormData - Post form data client.
+ *
+ * @returns {SubmitFormDataThunkFactory} Thunk factory.
+ */
+
+const submitFormData =
+  (postFormData: PostFormSchema) =>
```
boliveira, line 31: Same as https://github.com/Farfetch/blackout/pull/104/files#r778066035

packages/redux/src/forms/types/action.types.ts
```diff
@@ -0,0 +1,51 @@
+import type * as actionTypes from '../actionTypes';
```
boliveira, line 1: You do not need the `type` keyword here as you are not importing types.

packages/redux/src/forms/types/state.types.ts
```diff
@@ -0,0 +1,15 @@
+import type { CombinedState } from 'redux';
+import type { Error } from '@farfetch/blackout-client/types';
+import type { FormSchema } from '@farfetch/blackout-client/forms/types';
+
+export type State = CombinedState<{
```
boliveira, line 5: This state has several properties that are typed as an object that is indexed with a number key but these properties are indexed with the schemaCode property which is a string. So, I would change the type to string and I would replace `x:` with `schemaCode:`

packages/redux/src/forms/types/action.types.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+interface FetchFormSchemaFailureAction extends Action {
+  type: typeof actionTypes.FETCH_FORM_SCHEMA_FAILURE;
+  payload: { error: Error };
+  meta: { schemaCode: string };
+}
+
+/** Actions dispatched when the fetch form schema request is made. */
+export type FetchFormSchemaAction =
+  | FetchFormSchemaRequestAction
```
boliveira, line 23: Additional `|` there

packages/redux/src/forms/types/action.types.ts
```diff
@@ -0,0 +1,51 @@
...
+  meta: { schemaCode: string; data: any };
+}
+
+/** Actions dispatched when submits form schema request is made. */
+export type SubmitFormSchemaAction =
+  | SubmitFormSchemaRequestAction
+  | SubmitFormSchemaSuccessAction
+  | SubmitFormSchemaFailureAction;
+
+/** Actions dispatched when the reset size guides is called. */
```
boliveira, line 48: Actions dispatched when the reset `form schema` is called.

packages/redux/src/forms/types/action.types.ts
boliveira: It was actually correct this way, my mistake.

packages/client/src/forms/types/getFormSchema.types.ts
```diff
@@ -0,0 +1,8 @@
+import type { FormSchema } from './formSchema.types';
+import type { FormSchemaQuery } from './formSchemaQuery.types';
+
+export type GetFormSchema = (
+  schemaCode: string,
+  query?: FormSchemaQuery,
+  config?: Record<string, unknown>,
```
joaoprcosta, line 7: Missing Config type available on global types.

packages/client/src/forms/types/postFormSchema.types.ts
```diff
@@ -0,0 +1,7 @@
+import type { SubmitFormSchema } from './submitFormSchema.types';
+
+export type PostFormSchema = (
+  schemaCode: string,
+  data?: unknown,
+  config?: Record<string, unknown>,
```
joaoprcosta, line 6: Also here.

packages/redux/src/forms/actions/factories/fetchFormSchemaFactory.ts
```diff
@@ -52,7 +55,7 @@ export default getFormSchema =>
       });
 
       return result;
-    } catch (error) {
+    } catch (error: any) {
```
joaoprcosta, line 58: You have Error type in global types in the root of client package.

boliveira: Thanks! However, as Javascript does not support multiple catch clauses, we cannot use `Error` here (which also could be misleading typescript because the actual error thrown might not be an `Error` but something else that was thrown somewhere in code that is not ours). This raises a good question because we are not checking if the error is actually an Error and our error reducers are typed as returning an Error always. We should discuss it further internally but right now, let's leave the `any`. What do you think?

boliveira (approved):

ruifcnunes (approved):

packages/redux/src/forms/actions/factories/fetchFormSchemaFactory.ts
joaoprcosta: At least I think we should do what others are doing typing this error, maintaining the consistency between areas (except if this action is different than others). If you think the global approach it's not ok, an issue should be raised or we should talk about this in some sync.

boliveira: Agreed. However, when we migrate to typescript 4.0 we will have an error if the catch variable is not typed as any because it would be typed as `unknown` which is correct. I will bring that topic in a sync for us to discuss. In the meanwhile, @danielbento92, can you revert this to have the error variable not being explicitly typed as any?

joaoprcosta (approved):

ruifcnunes (approved):

PR #112, opened 2022-01-09
feat(redux): reset returns data on logout

jpklzm:
## Description

This resets the returns data (state and entities) when the user proceeds with a logout action.

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/returns/reducer.js
+++ b/packages/redux/src/returns/reducer.js
@@ -6,6 +6,7 @@
 
 import * as actionTypes from './actionTypes';
 import { combineReducers } from 'redux';
+import { LOGOUT_SUCCESS } from '../authentication/actionTypes';
 import { reducerFactory } from '../helpers';
 
 const INITIAL_STATE = {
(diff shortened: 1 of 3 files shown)


packages/redux/src/returns/reducer.js
```diff
@@ -56,6 +59,8 @@ const id = (state = INITIAL_STATE.id, action = {}) => {
       return action.payload.result;
     case actionTypes.RESET_RETURN:
       return INITIAL_STATE.id;
+    case LOGOUT_SUCCESS:
+      return INITIAL_STATE.id;
```
jhairabreu, line 63: this could be added to the previous case instead of adding a new one with the same info, i.e:

```
case actionTypes.RESET_RETURN
case LOGOUT_SUCCESS:
   INITIAL_STATE.id;
```

do you agree?

jhairabreu (changes requested):

packages/redux/src/returns/reducer.js
jpklzm: Make sense, I´ll change it

jhairabreu (approved):
Looks good to me 👍

andreecosta (approved):

pedro-gbf (approved):

PR #113, opened 2022-01-10
feat(client|redux): convert recently viewed area to ts

danielbento92:
## Description

- Conversion to typescript:
 - Client/recentlyViewed 
 - Redux/recentlyViewed

## Checklist

## Disclaimer


packages/redux/src/recentlyViewed/__tests__/reducer.test.ts
```diff
@@ -2,21 +2,22 @@ import * as fromReducer from '../reducer';
...
+} from 'tests/__fixtures__/recentlyViewed';
 import reducer, { actionTypes } from '..';
 
 let initialState;
+const randomAction = { type: 'this_is_a_random_action' };
 
 describe('Recently Viewed reducer', () => {
   beforeEach(() => {
-    initialState = reducer();
+    initialState = fromReducer.INITIAL_STATE;
```
boliveira, line 13: Question: Why was this change necessary?

packages/redux/src/recentlyViewed/actions/__tests__/fetchRecentlyViewedProducts.test.ts
```diff
@@ -12,10 +14,10 @@ jest.mock('@farfetch/blackout-client/recentlyViewed', () => {
   };
 });
 
-const mockRecentlyViewedStore = (state = {}) =>
+const mockRecentlyViewedStore = (state: StoreState = {}) =>
   mockStore(
     {
-      recentlyViewed: reducer(),
+      recentlyViewed: INITIAL_STATE,
```
boliveira, line 20: Same as https://github.com/Farfetch/blackout/pull/113/files#r783166100

packages/client/src/recentlyViewed/types/recentlyViewedProducts.types.ts
```diff
@@ -0,0 +1,17 @@
+export type RecentlyViewedProducts = RecentlyViewedProductsPaginationData &
+  RecentlyViewedProductsEntriesData;
+
+export type RecentlyViewedProductsPaginationData = {
+  number?: number;
```
boliveira, line 5: Question: Is there any case when these options are not available? I did a little test here and they seem to be always defined.

packages/redux/src/recentlyViewed/actions/factories/fetchRecentlyViewedProductsFactory.ts
```diff
@@ -0,0 +1,60 @@
...
+import type { StoreState } from '../../../types';
+
+/**
+ * Method to create a thunk factory configured with the specified client fo fetch recently viewed products.
+ *
+ * @param getRecentlyViewedProducts - Get recently viewed products client.
+ *
+ * @returns Thunk factory.
+ */
+
```
boliveira, line 19: Can you remove this empty line, please?

packages/redux/src/recentlyViewed/__tests__/reducer.test.ts
danielbento92: As I checked on other files it's used two approaches for this scenario. I found that it's being frequently changing the initialState to reducer's Initial state const, or in a few cases, uses a reducer with InitialState as a parameter. But the first approach is mostly used instead of the second one.

packages/redux/src/recentlyViewed/actions/factories/fetchRecentlyViewedProductsFactory.ts
```diff
@@ -0,0 +1,60 @@
...
+const fetchRecentlyViewedProducts: FetchRecentlyViewedProductsFactory<GetRecentlyViewedProducts> =
+
```
boliveira, line 21: Same as https://github.com/Farfetch/blackout/pull/113/files#r783181459

packages/redux/src/recentlyViewed/actions/factories/removeRecentlyViewedProductFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { RemoveRecentlyViewedProductsFactory } from './types';
+
+/**
+ * Method to create a thunk factory configured with the specified client for deleting a recently viewed product, both locally and on the server.
+ *
+ * @param deleteRecentlyViewedProduct - Delete recently viewed client.
+ *
+ * @returns Thunk factory.
+ */
+
```
boliveira, line 14: Same as https://github.com/Farfetch/blackout/pull/113/files#r783181459

packages/redux/src/recentlyViewed/actions/factories/removeRecentlyViewedProductFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+const removeRecentlyViewedProductFactory: RemoveRecentlyViewedProductsFactory<DeleteRecentlyViewedProducts> =
+
```
boliveira, line 16: Same as https://github.com/Farfetch/blackout/pull/113/files#r783181459

packages/client/src/recentlyViewed/types/deleteRecentlyViewedProducts.types.ts
```diff
@@ -0,0 +1,6 @@
+import type { Config } from '../../types';
+
+export type DeleteRecentlyViewedProducts = (
+  id: number,
+  config?: Config,
+) => Promise<unknown>;
```
boliveira, line 6: Question: Is this unknown? At least, it should return the request status, I believe.

packages/redux/src/recentlyViewed/actions/factories/removeRecentlyViewedProductFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+    deleteRecentlyViewedProduct =>
+    (productId, config) =>
+    async (
+      dispatch: Dispatch<RemoveRecentlyViewedProductAction>,
+    ): Promise<void> => {
```
boliveira, line 21: This void type does not match to the unknown type defined for the return value of the deleteRecentlyViewedProduct argument.

packages/redux/src/recentlyViewed/actions/factories/removeRecentlyViewedProductFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+      dispatch({
```
boliveira, line 22: Keep the dispatches inside the try / catch as well so that if an error is thrown while executing it, the store state can be updated with the error.

packages/client/src/recentlyViewed/types/getRecentlyViewedProducts.types.ts
```diff
@@ -0,0 +1,12 @@
+import type { Config } from '../../types';
+import type { RecentlyViewedProducts } from './recentlyViewedProducts.types';
+
+export type GetRecentlyViewedProductsQuery = {
+  page: number;
+  pageSize: number;
```
joaoprcosta, line 6: Because you set default values for each, I think it's better to type them.
Try this but I'm not sure if this is the way.

```suggestion
  page: number = 1;
  pageSize: number = 10;
```

packages/redux/src/recentlyViewed/__tests__/reducer.test.ts
boliveira: I understand but I think that the other way was better, because calling the reducer with an empty state should return the initial state by default. Imagine if tomorrow we change the reducer logic to return something else that is not exactly from the `INITIAL_STATE` variable. 

However, you can let it this way, as it is not critical.

packages/client/src/recentlyViewed/types/getRecentlyViewedProducts.types.ts
boliveira: In this case, these parameters are required. What is being done is if the query parameter is not sent to the getRecentlyViewedProducts client, a default is assumed that contains both the page and pageSize parameters set to a default value.  

```javascript
const getRecentlyViewedProducts: GetRecentlyViewedProducts = (
  query = { page: 1, pageSize: 10 },
  config,
)
```
And the type `GetRecentlyViewedProducts` reflects that as query is optional: 

```javascript
export type GetRecentlyViewedProducts = (
  query?: GetRecentlyViewedProductsQuery,
  config?: Config,
) => Promise<RecentlyViewedProducts>;
```

packages/redux/src/recentlyViewed/actions/factories/fetchRecentlyViewedProductsFactory.ts
danielbento92: In this case, I tried to remove but when format on saving, adds this line automatically.

packages/redux/src/recentlyViewed/actions/factories/removeRecentlyViewedProductFactory.ts
danielbento92: Same case as previous one, when format on saving, adds this line automatically

packages/client/src/recentlyViewed/types/deleteRecentlyViewedProducts.types.ts
danielbento92: This should be void, because as I check we don't get anything from response, only request status matters.

packages/client/src/recentlyViewed/getRecentlyViewedProducts.ts
```diff
@@ -26,11 +15,9 @@ import join from 'proper-url-join';
  *
  * @returns {Promise} Promise that will be resolved when the call to the endpoint finishes.
```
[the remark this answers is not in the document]
danielbento92: Source updated, can you check again.

packages/redux/src/recentlyViewed/actions/__tests__/fetchRecentlyViewedProducts.test.ts
danielbento92: Source updated, can you check again.

packages/client/src/recentlyViewed/types/recentlyViewedProducts.types.ts
danielbento92: Source updated, can you check again.

packages/client/src/recentlyViewed/types/recentlyViewedProducts.types.ts
```diff
@@ -0,0 +1,17 @@
...
+  totalPages?: number;
+  totalItems?: number;
+};
+
+export type RecentlyViewedProductsEntriesData = {
+  entries: RecentlyViewedProductsEntriesItem[];
+};
+
+export type RecentlyViewedProductsEntriesItem = {
+  productId?: number;
```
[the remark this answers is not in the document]
danielbento92: Source updated, can you check again.

packages/redux/src/recentlyViewed/actions/factories/fetchRecentlyViewedProductsFactory.ts
danielbento92: Source updated, can you check again.

packages/redux/src/recentlyViewed/actions/factories/removeRecentlyViewedProductFactory.ts
danielbento92: Source updated, can you check again.

packages/client/src/recentlyViewed/types/deleteRecentlyViewedProducts.types.ts
boliveira: Actually disregard this, you are right, it should be Promise< void > because the .then is returning the data field only and not all the response object.

packages/redux/src/recentlyViewed/actions/__tests__/fetchRecentlyViewedProducts.test.ts
boliveira: It actually should be this:

```javascript
recentlyViewed: reducer(undefined, mockAction),
```

as reducers should return the initial state when the state parameter is undefined.

packages/redux/src/recentlyViewed/actions/factories/removeRecentlyViewedProductFactory.ts
danielbento92: Can you check again. I update Type from Promise<unknown> to Promise<void>

boliveira (approved):

packages/redux/src/recentlyViewed/actions/__tests__/fetchRecentlyViewedProducts.test.ts
danielbento92: Source updated, can you check again.

joaoprcosta (approved):

PR #118, opened 2022-01-11
chore: add support for typescript in test files

boliveira:
## Description

- This adds support for using typescript in test files
while keeping the tests from transpiling when using yarn types command.
- Removed 'redux/tests' alias to be consistent with the other packages.

## Checklist

## Disclaimer

### DIFF
--- a/configs/typescript/tsconfig.basic.json
+++ b/configs/typescript/tsconfig.basic.json
@@ -1,23 +1,16 @@
 {
   "compilerOptions": {
-    "target": "ES2015" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
-    "module": "ES2020" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
-    "lib": [
-      "esnext",
-      "dom"
-    ] /* Specify library files to be included in the compilation. */,
-    "jsx": "preserve" /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */,
-    "isolatedModules": true /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */,
-    "strict": true /* Enable all strict type-checking options. */,
-    "moduleResolution": "node" /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */,
-    "noUncheckedIndexedAccess": true /* Will add undefined to any un-declared field in the type. */,
-    "allowSyntheticDefaultImports": true /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */,
-    "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */,
-    "resolveJsonModule": true /* Enable importing .json files." */,
-    "importsNotUsedAsValues": "error" /* Specify emit/checking behavior for imports that are only used for types */,
-    "typeRoots": [
-      "../node_modules/@types",
-      "../types"
-    ] /* List of folders to include type definitions from. */
+    "target": "ES2015",
+    "module": "ES2020",
+    "lib": ["esnext", "dom"],
+    "jsx": "preserve",
+    "isolatedModules": true,
+    "strict": true,
+    "moduleResolution": "node",
+    "noUncheckedIndexedAccess": true,
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true,
+    "importsNotUsedAsValues": "error"
   }
 }
--- a/packages/analytics/tsconfig.build.json
+++ b/packages/analytics/tsconfig.build.json
@@ -1,0 +1,5 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": ["**/__tests__/*", "**/__fixtures__/*"],
+  "include": ["src", "../../types"]
+}
--- a/packages/analytics/tsconfig.json
+++ b/packages/analytics/tsconfig.json
@@ -1,15 +1,8 @@
 {
+  "extends": "../../configs/typescript/tsconfig.package-build.json",
   "compilerOptions": {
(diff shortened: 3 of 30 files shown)


configs/typescript/tsconfig.basic.json
```diff
@@ -14,10 +14,6 @@
     "allowSyntheticDefaultImports": true /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */,
     "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */,
     "resolveJsonModule": true /* Enable importing .json files." */,
-    "importsNotUsedAsValues": "error" /* Specify emit/checking behavior for imports that are only used for types */,
-    "typeRoots": [
-      "../node_modules/@types",
-      "../types"
-    ] /* List of folders to include type definitions from. */
+    "importsNotUsedAsValues": "error" /* Specify emit/checking behavior for imports that are only used for types */
```
helderburato, line 17: Is this really necessary to keep an explanation of every line in the configuration file? I think when in doubt about some rules we can read the [TypeScript docs](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html). What do you think about it?

packages/analytics/tsconfig.build.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "./tsconfig.json",
```
helderburato, line 2: Considering we have some `tsconfig.*.json` files, it could be nice to have a base file, something like `tsconfig.base.json` just to be more readable what is the basis for every configuration.

packages/analytics/tsconfig.json
```diff
@@ -1,15 +1,11 @@
 {
+  "extends": "../../configs/typescript/tsconfig.package-build.json",
```
helderburato, line 2: What's the difference between `tsconfig.build.json` and `tsconfig.package-build.json`? Maybe we can describe these configurations in the README or another place.

packages/redux/src/analytics/middlewares/__tests__/bag.test.js
```diff
@@ -9,7 +9,7 @@ import {
   mockBrandId,
   mockCategoryId,
   mockStore,
-} from 'redux/tests';
+} from '../../../../tests';
```
helderburato, line 12: Consider configuring an alias for cases like this.

helderburato (changes requested):
I think nothing is blocking, but the `tests` aliases could be helpful for these cases.

configs/typescript/tsconfig.basic.json
boliveira: Yes, I agree, I just copy/pasted the contents from the other existing files and as I did not think it would hurt to have the comments there, I kept it but I believe now it is better to remove them. Let's see if the others also agree as well, but my vote is to remove them. Thanks!

packages/analytics/tsconfig.build.json
boliveira: Having a tsconfig.json was intentional. This is because when you edit code in Visual Studio Code, it will detect files named only as tsconfig.json (AFAIK) and will correctly load the configurations so that tsserver can emit errors/warnings for your code when you are editing it. 

That repository you mentioned I believe is for another purpose. In this case, we need to have a tsconfig.json here not only because of the editing experience but also because of generating the correct output when building.

What do you think?

packages/analytics/tsconfig.json
boliveira: The purpose of those files are to be used only when running `yarn types` command, which will be used by the pipeline when we migrate the codebase fully to typescript. This is because the default tsconfig.json enables typings of test files and we do not want to have them outputted when we generate a new version of the package.

I was thinking of naming them exactly as `tsconfig.package-build.json` but the reason I didn't pick that name is that it might get more difficult to navigate in your IDE when you have many files with the same name opened.

What do you think? Do you have a better naming strategy?

packages/redux/src/analytics/middlewares/__tests__/bag.test.js
boliveira: There was an alias for that (It was redux/tests) but I removed it because the other packages were doing the same thing as well, so I was trying to be consistent in this case. What do you think?

packages/analytics/tsconfig.build.json
boliveira: More info about this: https://github.com/microsoft/vscode/issues/12463

helderburato: It's just a suggestion, I'm not a VSCode user, and for me (and maybe other people) it doesn't give any benefits.

IMHO it doesn't hurt to keep this filename, and it probably can be configured in the VSCode to use other files as the basis.

packages/redux/src/analytics/middlewares/__tests__/bag.test.js
helderburato: No problem, it was a suggestion just to improve the readability, in my perspective `../../../../tests` seems weird.

packages/analytics/tsconfig.json
helderburato: Naming is always a problem 🤣  to be honest, I don't have a better suggestion I was trying to understand the purpose of this file.

helderburato (approved):

packages/analytics/tsconfig.build.json
boliveira: Unfortunately tsserver does not allow that configuration ([microsoft/vscode#12463](microsoft/vscode#12463)), we have to use tsconfig.json for VSCode to detect it. (Btw tsserver is the typescript language server that the IDEs can use to add support for typescript in editors, so I believe this might affect any Editor that uses tsserver underneath to provide support for typescript).

configs/typescript/tsconfig.basic.json
nelsonleite: Yeah, I'd also say to remove them. If. there is any doubt, we go to the documentation. The same applies to the properties that are not there 😄 

_(applicable to all config files)_

boliveira: Removed! Thanks guys!

nelsonleite (approved):

PR #120, opened 2022-01-12
feat(react): add useProductDetails hook

ricardojmf:
## Description

This adds a useProductDetails hook that includes all logic associated with the product.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
+++ b/packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
@@ -1,0 +1,75 @@
+import { fetchProductDetails } from '@farfetch/blackout-redux/products';
+import { mockProductId, mockProductsState } from 'tests/__fixtures__/products';
+import { mockStore } from '../../../../tests/helpers';
+import { Provider } from 'react-redux';
+import { renderHook } from '@testing-library/react-hooks';
+import React from 'react';
+import useProductDetails from '../useProductDetails';
+
+jest.mock('@farfetch/blackout-redux/products', () => ({
+  ...jest.requireActual('@farfetch/blackout-redux/products'),
+  fetchProductDetails: jest.fn(() => ({ type: 'get' })),
+}));
+
+const getRenderedHook = (props, state = mockProductsState) => {
+  return renderHook(() => useProductDetails(props), {
+    wrapper: ({ children }) => (
+      <Provider store={mockStore(state)}>{children}</Provider>
+    ),
+  });
+};
+
+describe('useProductDetails', () => {
+  beforeEach(jest.clearAllMocks);
+
+  describe('when it has a product id', () => {
+    it('should return values correctly with initial state', async () => {
+      const { result } = getRenderedHook(mockProductId);
+
+      expect(result.current).toStrictEqual({
+        breadcrumbs: expect.any(Array),
+        error: expect.any(Object),
+        groupedEntries: undefined,
+        isDuplicated: expect.any(Boolean),
+        isFetched: expect.any(Boolean),
+        isHydrated: expect.any(Boolean),
+        isInBag: expect.any(Boolean),
+        isLoading: expect.any(Boolean),
+        isOneSize: expect.any(Boolean),
+        isOutOfStock: expect.any(Boolean),
+        labelsPrioritized: expect.any(Array),
+        product: expect.any(Object),
+        promotions: expect.any(Array),
+        availableSizes: expect.any(Array),
+      });
+    });
+  });
+
+  describe('when the product has not been fetched', () => {
+    it('should fetch the product', () => {
+      const storeState = {
+        ...mockProductsState,
+        products: {
+          ...mockProductsState.products,
+          details: {
+            ...mockProductsState.products.details,
+            isHydrated: {},
+            isLoading: {},
+          },
+        },
+      };
+
+      getRenderedHook(mockProductId, storeState);
+
+      expect(fetchProductDetails).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('when the product is already fetched', () => {
+    it('should not fetch the product', () => {
+      getRenderedHook(mockProductId);
+
+      expect(fetchProductDetails).not.toHaveBeenCalled();
+    });
+  });
+});
--- a/packages/react/src/products/hooks/types/useProductDetails.ts
+++ b/packages/react/src/products/hooks/types/useProductDetails.ts
@@ -1,0 +1,27 @@
+import type {
+  BreadCrumb,
+  Product,
+} from '@farfetch/blackout-client/src/products/types';
+import type { Error } from '@farfetch/blackout-client/src/types';
+import type {
+  GroupedEntriesAdapted,
+  ProductEntity,
+} from '@farfetch/blackout-redux/src/entities/types';
+import type { SizesAdapted } from '@farfetch/blackout-client/src/helpers/adapters/types';
+
+export type UseProductDetails = (id: number) => {
+  availableSizes: SizesAdapted;
+  breadcrumbs: BreadCrumb[] | undefined;
+  error: Error | undefined;
+  groupedEntries: GroupedEntriesAdapted;
+  isDuplicated: boolean | undefined;
+  isFetched: boolean;
+  isHydrated: boolean | undefined;
+  isInBag: boolean;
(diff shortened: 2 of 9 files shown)


mindouro (approved):

packages/react/src/products/hooks/useProductDetails.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Provides state access for dealing with product details business logic.
+ * Also it fetches product details if there are none.
+ *
+ * @param {number} id - Product id to work on.
+ *
+ * @memberof module:products/hooks
+ * @returns {object} All the state needed to manage any product details information.
+ */
+const useProductDetails: UseProductDetails = id => {
+  const getProductDetails = useAction(fetchProductDetails);
```
mindouro, line 43: To not mistake this with the selector, what do you think of keeping the `fetchProductDetails` name?

mindouro (changes requested):

mindouro (approved):

anapim (approved):

packages/redux/src/entities/selectors/products.ts
```diff
@@ -96,7 +96,7 @@ export const getProductLabelsByPriority = createSelector(
   [
     (state: StoreState, productId: ProductEntity['id']) =>
       getProduct(state, productId),
-    (state: StoreState, productId: ProductEntity['id'], sortOrder: string) =>
+    (state: StoreState, productId: ProductEntity['id'], sortOrder = 'asc') =>
```
boliveira, line 99: Is it possible to type the sortOrder parameter? From the code I can only see it has a default value 'asc' but I cannot know which values it can accept.

packages/react/src/products/hooks/useProductDetails.ts
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * The available sizes for the given product.
+     *
+     * @type {Array}
+     */
+    availableSizes,
+    /**
+     * Product breadcrumbs.
+     *
+     * @type {Array|undefined}
```
boliveira, line 95: These types are not exactly matching the ones defined by `UseProductDetails` type.
Also, I believe the @type tag is unnecessary as the type information is already available in typescript, so you can remove the tag (not the comment).

packages/redux/src/entities/selectors/products.ts
ricardojmf: Sure nice catch!

packages/react/src/products/hooks/useProductDetails.ts
ricardojmf: Yes they are not exactly matching but this is for jsdoc purposes not typescript

boliveira: I understand but when you have typescript, typings should not be a concern of the jsdoc. We will have to migrate to tsdoc in the future which looks something like this: https://tsdoc.org/

Anyway, at least the same types could be used between jsdoc and `UseProductDetails` type, so there aren't any mismatches if someone changes the type in Typescript and forgets to change the type in the jsdoc.

For the breadcrumbs property it could be:

```javascript
/**
     * Product breadcrumbs.
     *
     * @type {BreadCrumb[] | undefined}
```

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  ({ children }) => {
+    const finalStoreState = {
+      ...mockProductsState,
+      ...storeState,
+    };
+
+    return <Provider store={mockStore(finalStoreState)}>{children}</Provider>;
+  };
+
+const renderHookWithRedux = (props, storeState = undefined) => {
```
nelsonleite, line 30: What about having a `mockProductDetailsState` as the default of the `storeState`? Not only it's consistent with the other hooks "strategy", but it also gets simpler and shorter (you stop needing `ReduxProviderWrapper` and an assembled `finalStoreState`).

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return renderHook(() => useProductDetails(props), {
+    wrapper: ReduxProviderWrapper(storeState),
+  });
+};
+
+describe('useProductDetails', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
```
nelsonleite, line 39:
```suggestion
  beforeEach(jest.clearAllMocks);
```

Kind of a nitpick 🙂

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  describe('when it has a product id', () => {
+    it('should return the correct values', async () => {
```
nelsonleite, line 42: Shouldn't this be "should return values correctly with initial state", using the previously commented `mockProductDetailsState`, instead of creating a state just for this test? 🤔

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
nelsonleite: And for the sake of consistency, I'd call this `getRenderedHook` instead of `renderHookWithRedux`.

packages/react/src/products/hooks/useProductDetails.ts
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Product grouped entries.
+     *
+     * @type {Array|undefined}
+     */
+    groupedEntries,
+    /**
+     * Whether the product is duplicated.
+     *
+     * @type {boolean}
```
nelsonleite, line 113: You should review all of these types, because in many cases it's `boolean | undefined` (that discussion we already had 😅 )

packages/react/src/products/hooks/useProductDetails.ts
ricardojmf: Its a good idea to have a more integrated documentation with typescript, but I think that its better to change documentation after changing the technology :)

packages/react/src/products/hooks/types/useProductDetails.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { Error } from '@farfetch/blackout-client/src/types';
+import type {
+  GroupedEntriesAdapted,
+  ProductEntity,
+} from '@farfetch/blackout-redux/src/entities/types';
+import type { SizesAdapted } from '@farfetch/blackout-client/src/helpers/adapters/types';
+
+export type UseProductDetails = (id: number) => {
+  availableSizes: SizesAdapted;
+  breadcrumbs: BreadCrumb[] | undefined;
```
joaoprcosta, line 14: What do you think?

```suggestion
  breadcrumbs?: BreadCrumb[];
```

nelsonleite (changes requested):

packages/react/src/products/hooks/useProductDetails.ts
boliveira: Ok, I will close this but keep in mind that the types between jsdoc and `UseProductDetails` can be out-of-sync.

packages/react/src/products/hooks/types/useProductDetails.ts
ricardojmf: I prefer the `BreadCrumb[] | undefined` approach. Although the represent the same "data validity", in terms of presentation and developer experience they are different because the first approach forces the return object to have the `breadcrumbs` key to appear in the object and the other one does not :)

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+describe('useProductDetails', () => {
+  beforeEach(jest.clearAllMocks);
+
+  describe('when it has a product id', () => {
+    it('should return values correctly with initial state', async () => {
+      const { result } = getRenderedHook(mockProductId);
+
+      expect(result.current).toStrictEqual({
+        breadcrumbs: expect.any(Array),
+        error: expect.any(String),
```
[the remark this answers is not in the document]
ricardojmf: Changed 😄

nelsonleite (approved):

joaoprcosta (approved):

PR #126, opened 2022-01-14
feat: add `addWishlistItem` to `useWishlist` hook

mindouro:
## Description
This adds the `addWishlistItem` method, responsible for adding an item to the wishlist, to the
`useWishlist` hook.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/wishlists/hooks/types/useWishlist.ts
+++ b/packages/react/src/wishlists/hooks/types/useWishlist.ts
@@ -1,10 +1,21 @@
-import type { State } from '@farfetch/blackout-redux/wishlists/types';
+import type { AxiosRequestConfig } from 'axios';
+import type {
+  PostWishlistItemActionData,
+  State,
+} from '@farfetch/blackout-redux/wishlists/types';
 import type { Wishlist } from '@farfetch/blackout-client/wishlists/types';
 import type { WishlistItemHydrated } from '@farfetch/blackout-redux/entities/types';
 
 export type UseWishlist = () => {
   error: State['error'] | undefined;
-  fetchWishlist: () => Promise<Wishlist | undefined>;
+  fetchWishlist: (
+    wishlistId: number,
+    config?: AxiosRequestConfig,
+  ) => Promise<Wishlist | undefined>;
+  addWishlistItem: (
+    data: PostWishlistItemActionData,
+    config?: AxiosRequestConfig,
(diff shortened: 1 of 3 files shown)


packages/react/src/wishlists/hooks/types/useWishlist.ts
```diff
@@ -5,6 +5,7 @@ import type { WishlistItemHydrated } from '@farfetch/blackout-redux/entities/typ
 export type UseWishlist = () => {
   error: State['error'] | undefined;
   fetchWishlist: () => Promise<Wishlist | undefined>;
+  addWishlistItem: () => Promise<Wishlist | undefined>;
```
nelsonleite, line 8: I believe you're missing the payload type here (it has the same type as the respective action, with the `data` and `config`)

And since you are here, is it out of hand to double-check if [this](https://github.com/Farfetch/blackout/blob/next/packages/react/src/wishlists/hooks/types/useWishlistSets.ts#L13) has the same problem? At least for the `config` part 😬

nelsonleite (changes requested):

packages/react/src/wishlists/hooks/types/useWishlist.ts
anapim: I had the same doubt about `config` but we don't seem to type it in any other action (they all receive `config` as a parameter). Should we add `config` to all action types? 🤔

catarinacouto (approved):

packages/react/src/wishlists/hooks/types/useWishlist.ts
nelsonleite: True, in fact, we don't type it anywhere 😅 

Should we open an issue to tackle this in all hooks?

mindouro: Good catch! I added both `data` and `config` to the action's arguments. Not having the `config` anywhere is indeed a problem, thanks for raising awareness for that problem. 

We're missing it in the hooks and in the action presets, I think we should discuss this in order to get a viable solution and then open an issue, what do you think? 😃

nelsonleite: Yup, we'll open an issue after a brief discussion 😄

nelsonleite (approved):

anapim (approved):

ricardojmf (approved):

Issue #18, opened 2021-11-23
Replace `moxios` with an alternative


nelsonleite:
## Is your proposal related to a problem?

`moxios` seems to be incompatible with `axios` after version 0.19.1, as reported [here](https://github.com/axios/axios/issues/2650). Besides, `moxios` latest release is 4+ years old.

The seems to be not in `moxios` actually but rather in `axios` itself and the problem was that after `axios` version 0.18, the code that was concatenating the `baseURL` with the URL of the request was done before much earlier and now it is done on each `axios` adapter class (There is an adapter for XHR requests and HTTP requests). Now, `moxios` replaces these adapters with its own `mockAdapter`, so the code that built the full URL was now not being executed.

## Describe the solution you'd like

**Possible options**

- [nock](https://github.com/nock/nock)
- [axios-mock-adapter](https://github.com/ctimmerm/axios-mock-adapter)
- [msw](https://github.com/mswjs/msw)

Here is a [comparative](https://www.npmtrends.com/fetch-mock-vs-moxios-vs-nock) between them.

The team chooses MSW.

- [x] addresses
- [ ] ~~analyticsService~~
- [x] authentication
- [x] bags
- [x] brands
- [x] categories
- [x] checkout
- [x] contents
- [x] designers
- [x] forms
- [x] language
- [x] locale
- [x] loyalty
- [x] merchantsLocations
- [ ] ~~oAuth~~
- [x] omnitracking
- [x] orders
- [x] payments
- [x] products
- [x] promotionEvaluations
- [x] recentlyViewed
- [x] recommendations
- [x] returns
- [x] search
- [x] settings
- [x] siteFeatures
- [x] sizeGuides
- [x] sizeScales
- [x] staffMembers
- [x] subscriptions
- [x] users
- [x] wishlists

jhairabreu:
closed by https://github.com/Farfetch/blackout/pull/347

PR #130, opened 2022-01-17
feat(client): add mock service worker contents

joaoprcosta:
## Description

- Remove moxios inside contents
- add msw to fixtures

Refs #18 

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/client/src/contents/__fixtures__/commercepages.fixtures.js
+++ b/packages/client/src/contents/__fixtures__/commercepages.fixtures.js
@@ -1,11 +1,6 @@
-import join from 'proper-url-join';
-import moxios from 'moxios';
+import { rest } from 'msw';
 
-/**
- * @typedef {object} RequestParams
- * @property {object} query - Query to request.
- * @property {Array} [response] - Contents payload.
- */
+const path = '/api/content/v1/commercepages';
 
 /**
  * Response payloads.
@@ -13,36 +8,20 @@ import moxios from 'moxios';
 export default {
   get: {
     /**
-     * Success moxios request.
+     * Success msw request.
      *
-     * @param {RequestParams} params - Params to moxios request.
+     * @param {object} response - Commerce Pages payload.
      */
-    success: params => {
-      moxios.stubRequest(
-        join('/api/content/v1/commercepages', {
-          query: params.queryParams,
-        }),
-        {
-          response: params.response,
-          status: 200,
-        },
-      );
-    },
+    success: response =>
+      rest.get(path, async (req, res, ctx) =>
+        res(ctx.status(200), ctx.json(response)),
+      ),
     /**
-     * Failure moxios request.
-     *
-     * @param {RequestParams} params - Params to moxios request.
+     * Failure msw request.
      */
-    failure: params => {
-      moxios.stubRequest(
-        join('/api/content/v1/commercepages', {
-          query: params.queryParams,
-        }),
-        {
-          response: 'stub error',
-          status: 404,
-        },
-      );
-    },
+    failure: () =>
+      rest.get(path, async (req, res, ctx) =>
+        res(ctx.status(404), ctx.json({ message: 'stub error' })),
+      ),
   },
 };
--- a/packages/client/src/contents/__fixtures__/contentTypes.fixtures.js
+++ b/packages/client/src/contents/__fixtures__/contentTypes.fixtures.js
@@ -1,10 +1,4 @@
-import moxios from 'moxios';
-
-/**
- * @typedef {object} RequestParams
- * @property {string} spaceCode - The space the content belongs to (website|mobileapp|emailTool...).
- * @property {Array} [response] - Content types payload.
- */
+import { rest } from 'msw';
 
 /**
  * Response payloads.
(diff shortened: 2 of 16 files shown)


packages/client/src/contents/__fixtures__/commercepages.fixtures.js
```diff
@@ -1,48 +1,27 @@
...
-        }),
-        {
-          response: params.response,
-          status: 200,
-        },
-      );
-    },
+    success: response =>
+      rest.get(path, async (req, res, ctx) =>
+        res(ctx.status(200), ctx.json({ ...response })),
```
helderburato, line 17: I think you don't need to destructure `{ ...response }` you can just pass response as an argument to `ctx.join(...)`:
```js
ctx.join(response)
```

packages/client/src/contents/__fixtures__/contentTypes.fixtures.js
```diff
@@ -1,43 +1,29 @@
...
-          response: params.response,
-          status: 200,
-        },
-      );
-    },
+    success: response =>
+      rest.get(
+        '/api/content/v1/spaces/:spaceCode/contentTypes',
+        async (req, res, ctx) =>
+          res(ctx.status(200), ctx.json({ ...response })),
```
helderburato, line 17: Same as above.

packages/client/src/contents/__tests__/getCommercePages.test.js
```diff
@@ -1,18 +1,15 @@
...
 
 describe('getCommercePages()', () => {
   const expectedConfig = undefined;
 
   beforeEach(() => {
-    moxios.install(client);
     jest.clearAllMocks();
   });
 
-  afterEach(() => moxios.uninstall(client));
```
helderburato, line 14: I didn't tested, but I think you can call `resetHandlers()` here:
```js
afterEach(() => {
  mswServer.resetHandlers();
})
```

As mentioned in the [MSW docs](https://mswjs.io/docs/api/setup-server/reset-handlers#default-behavior):
_The .resetHandlers() function is useful as a clean up mechanism between multiple test suites that leverage runtime request handlers._

helderburato (approved):
Great work 🚀

tests/mswServer.js
```diff
@@ -0,0 +1,3 @@
+import { setupServer } from 'msw/node';
+
+export const mswServer = setupServer();
```
nelsonleite, line 3: I think I'd replace this for a default export, feels more natural to me when importing, since this is a single setup 🤔 

```js
const mswServer = setupServer();

export default mswServer;
```

tests/setupAfterEnv.js
```diff
@@ -2,3 +2,8 @@ import '@testing-library/jest-dom/extend-expect';
 import 'core-js/stable';
 import 'jest-axe/extend-expect';
 import 'regenerator-runtime/runtime';
+import { mswServer } from './mswServer';
+
+beforeAll(() => mswServer.listen());
```
nelsonleite, line 7: Won't this run everywhere, including places where this doesn't make sense, for example the react package?

Is it worth it, or is it useful anyway for every single test?

nelsonleite (changes requested):

mindouro (approved):

ricardojmf (approved):

packages/client/src/contents/__fixtures__/commercepages.fixtures.js
```diff
@@ -1,48 +1,27 @@
...
  * Response payloads.
  */
 export default {
   get: {
     /**
-     * Success moxios request.
+     * Success msw request.
      *
-     * @param {RequestParams} params - Params to moxios request.
+     * @param {Array} response - Commerce Pages payload.
```
boliveira, line 13: Response seems to be an object and not an array. Can you check if that is the expected type?

packages/client/src/contents/__fixtures__/contentTypes.fixtures.js
```diff
@@ -1,43 +1,29 @@
...
  * Response payloads.
  */
 export default {
   get: {
     /**
-     * Success moxios request.
+     * Success msw request.
      *
-     * @param {RequestParams} params - Params to moxios request.
+     * @param {Array} response - Content types payload.
```
boliveira, line 11: Same as https://github.com/Farfetch/blackout/pull/130/files#r786201319

packages/client/src/helpers/client/interceptors/authentication/__tests__/AxiosAuthenticationTokenManager.test.js
```diff
@@ -679,10 +679,9 @@ describe('AxiosAuthenticationTokenManager', () => {
     it('should throw an error if a request is performed before token manager is loaded', () => {
       tokenManagerInstance = createAndSetTokenManagerInstance(client);
 
-      expect(tokenManagerInstance.isLoaded).toBeFalsy();
-
-      expect.assertions(1);
+      expect.assertions(2);
```
boliveira, line 682: Question: Why was this necessary?

tests/setupAfterEnv.js
nelsonleite: Wild suggestion - we could have this isolated and in every test needed, we just import and use `mswServerSetup()` 🪄

packages/client/src/helpers/client/interceptors/authentication/__tests__/AxiosAuthenticationTokenManager.test.js
joaoprcosta: This test failed after msw config. But, if you check, there are two expects in this test. I'm curious how did the test pass before this configuration...

packages/client/src/contents/__fixtures__/commercepages.fixtures.js
joaoprcosta: Done

tests/mswServer.js
joaoprcosta: Done

tests/setupAfterEnv.js
joaoprcosta: Could you check please?

packages/client/src/contents/__fixtures__/commercepages.fixtures.js
joaoprcosta: Done.

packages/client/src/contents/__fixtures__/contentTypes.fixtures.js
joaoprcosta: Done.

packages/client/src/helpers/client/interceptors/authentication/__tests__/AxiosAuthenticationTokenManager.test.js
joaoprcosta: After the new config setup for msw, one assertion was only needed but this is strange because you have two expects in the same test.

boliveira: Interesting question. I dig a little deeper in jest to understand why the test was failing now when it was passing before and the thing is, the added `afterEach` call to setupAfterEnv.js will change the timings internally in jest, namely when it will process the assertions. Without the `afterEach` call it will process the results right before the call to expect(getUser()).rejects... which will reset the internal assertions number to 0 while with the `afterEach` call it will only do that after all expects are done.

TLDR; 
The test code was in fact incorrect, it should be `expect.assertions(2)` as you have done to avoid these timings problems in jest depending on the kind of tests being done and the afterEach blocks defined. Closing it... 

Thanks!

nelsonleite (approved):

PR #145, opened 2022-01-21
feat(redux): add isWishlistSetFetched selector

ricardojmf:
## Description

This adds a new isWishlistFetched selector

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/wishlists/selectors/__tests__/wishlistsSets.test.ts
+++ b/packages/redux/src/wishlists/selectors/__tests__/wishlistsSets.test.ts
@@ -87,6 +87,60 @@ describe('wishlists redux selectors', () => {
     });
   });
 
+  describe('isWishlistSetFetched()', () => {
+    it('should get true if the wishlist has been fetched and it is not loading', () => {
+      const newMockWishlistState = {
+        ...mockWishlistState,
+        wishlist: {
+          ...mockWishlistState.wishlist,
+          sets: {
+            ...mockWishlistState.wishlist.sets,
+            set: {
+              ...mockWishlistState.wishlist.sets.set,
+              isLoading: {
+                ...mockWishlistState.wishlist.sets.set.isLoading,
+                [mockWishlistSetId]: false,
+              },
+            },
+          },
+        },
+      };
+
+      expect(
+        selectors.isWishlistSetFetched(newMockWishlistState, mockWishlistSetId),
+      ).toBe(true);
+    });
+
+    it('should get false if it is loading', () => {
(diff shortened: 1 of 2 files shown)


packages/redux/src/wishlists/selectors/__tests__/wishlistsSets.test.ts
```diff
@@ -87,6 +87,100 @@ describe('wishlists redux selectors', () => {
...
+        },
+      };
+
+      expect(
+        selectors.isWishlistSetFetched(newMockWishlistState, mockWishlistSetId),
+      ).toBe(true);
+      expect(spyGetIsSetLoading).toHaveBeenCalledTimes(1);
+      expect(spyGetEntityById).toHaveBeenCalledTimes(1);
+    });
+    it('should get false if the there is no setId', () => {
```
mindouro, line 120: We usually have an empty line between `it`s to increase the separation between them. Could you add empty lines in here as well?

mindouro (approved):
💪

anapim (approved):

packages/redux/src/wishlists/selectors/wishlistsSets.ts
```diff
@@ -132,6 +132,24 @@ export const isWishlistSetLoading = (
...
+ * @param {object} state - Application state.
+ * @param {string} setId - Global identifier of the set in the wishlist.
+ *
+ * @returns {boolean} If a certain set has been fetched or not.
+ */
+export const isWishlistSetFetched = (
+  state: StoreState,
+  setId: WishlistSet['setId'],
+): boolean =>
+  !!setId &&
```
nelsonleite, line 149: Here I'd probably do something similar to other selectors like the fittings example

```js
getIsLoading(state.products.fittings).hasOwnProperty(id) && 
areProductFittingsLoading(state, id) === false;
```

Because in fact, your last verification (`!!getWishlistSet(state, setId)`) alone does that, if you can get the wishlist set, it was fetched. (Isn't it too much to check if something is fetched? Can't you have an empty object being returned for some reason, thus turning this a false positive? 🤔 )

I don't know, it's more a consistency thing... And for example, why do you need to verify `!!setId` if the argument is not optional? If its value is undefined for some reason, the `isWishlistSetLoading` already returns the correct value for that, you know what I mean?

nelsonleite (changes requested):

nelsonleite (approved):

anapim (approved):

mindouro (approved):

PR #152, opened 2022-01-26
feat(react): add `useProductsList` hook

mindouro:
## Description
This adds the `useProductsList` hook, with selectors and logic useful
in products lists.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/products/hooks/types/useProductsList.ts
+++ b/packages/react/src/products/hooks/types/useProductsList.ts
@@ -1,0 +1,37 @@
+import type { Error } from '@farfetch/blackout-client/src/types';
+import type {
+  ListingQuery,
+  SetQuery,
+} from '@farfetch/blackout-client/src/products/types';
+import type {
+  ProductEntity,
+  ProductsListEntity,
+} from '@farfetch/blackout-redux/src/entities/types';
+
+export type ProductsListTypes = {
+  LISTING: 'listing';
+  SET: 'set';
+};
+
+export type UseProductsListParams = {
+  query: ListingQuery | SetQuery;
+  setProductsListHash?: boolean;
+  slug: string;
+  type: ProductsListTypes['LISTING'] | ProductsListTypes['SET'];
+  useCache?: boolean;
+};
+
+export type UseProductsList = ({
+  query,
+  setProductsListHash,
+  slug,
+  type,
+  useCache,
+}: UseProductsListParams) => {
+  error: Error | undefined;
+  isFetched: boolean | undefined;
+  isLoading: boolean | undefined;
+  products: ProductEntity[];
+  productsListHash: string;
+  result: ProductsListEntity;
+};
--- a/packages/react/src/products/hooks/useProductsList.ts
+++ b/packages/react/src/products/hooks/useProductsList.ts
@@ -1,0 +1,135 @@
+/**
+ * Hook to provide all kinds of data for the business logic attached to a product list.
+ *
+ * @module useProductsList
+ * @category Products
+ * @subcategory Hooks
+ */
+import {
+  fetchListing,
+  fetchSet,
+  getProductsListError,
+  getProductsListProducts,
+  getProductsListResult,
+  isProductsListFetched,
+  isProductsListLoading,
+} from '@farfetch/blackout-redux/products';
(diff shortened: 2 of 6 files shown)


nelsonleite (approved):

packages/react/src/products/hooks/useProductsList.ts
```diff
@@ -0,0 +1,94 @@
+import {
```
anapim, line 1: I think you're missing some documentation up here 🙂 
```
/**
 * Hook to provide ...
 *
 * @module useProductsList
 * @category Products
 * @subcategory Hooks
 */
 ```

File: packages/react/src/products/hooks/useProductsList.ts [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+    slug,
+    useCache,
+  ]);
+
+  return {
+    error,
+    isLoading,
+    productsListHash,
+    products,
+    result,
```
anapim, line 90: And jsdoc here, or did we decide on not using it anymore?

anapim (changes requested):

packages/react/src/products/hooks/types/useProductsList.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type UseProductsListParams = {
+  query: ListingQuery | SetQuery;
+  setProductsListHash?: boolean;
+  slug: string;
+  type: ProductsListTypes['LISTING'] | ProductsListTypes['SET'];
+  useCache?: boolean;
+};
+
+export type UseProductsList = {
```
anapim, line 24:
```suggestion
export type UseProductsList = (params: UseProductsListParams) => {
```

And this way we simply use one type on our hook. I think this way we keep it consistent with other hooks.

packages/react/src/products/hooks/useProductsList.ts
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * @returns {Object} All the info needed to get listing results and information.
+ */
+const useProductsList = ({
+  query,
+  setProductsListHash = true,
+  slug,
+  type = PRODUCTS_LIST_TYPES.LISTING,
+  useCache = true,
+}: UseProductsListParams): UseProductsList => {
```
anapim, line 52: If you accept my previous suggestion we simplify this to:

```suggestion
const useProductsList: useProductsList = ({
  query,
  setProductsListHash = true,
  slug,
  type = PRODUCTS_LIST_TYPES.LISTING,
  useCache = true,
}) => {
```

anapim (changes requested):

packages/react/src/products/hooks/useProductsList.ts
mindouro: I totally forgot about the extra JsDoc we add here 🙈 fixed now!

packages/react/src/products/hooks/useProductsList.ts
mindouro: Added! 😄

packages/react/src/products/hooks/types/useProductsList.ts
mindouro: Changed :)

packages/react/src/products/hooks/useProductsList.ts
mindouro: I know eheh thanks! 😁

anapim (approved):

packages/react/src/products/hooks/useProductsList.ts
```diff
@@ -0,0 +1,125 @@
...
+  );
+  const result = useSelector(state =>
+    getProductsListResult(state, productsListHash),
+  );
+  const products = useSelector(state =>
+    getProductsListProducts(state, productsListHash),
+  );
+
+  useEffect(() => {
+    if (slug !== undefined && !products) {
```
ricardojmf, line 71: A small suggestion, what do you think of using this selector  `isProductsListFetched` for the "if" inside the useEffect, and also adding it to the return values of the hook :)?

mindouro: Sounds perfect, nice catch :) Changed now!

ricardojmf (approved):

PR #153, opened 2022-01-27
feat(redux): improve actions of 'users'

jpklzm:
## Description

Adds return statements to the `users` actions
Adds missing types to `users` actions

### Dependencies
None

## Checklist

## Disclaimer


packages/client/src/users/types/index.ts
```diff
@@ -7,3 +7,6 @@ export * from './postUserAttributes.types';
 export * from './putUserAttribute.types';
 export * from './patchUserAttribute.types';
 export * from './deleteUserAttribute.types';
+export * from './contact.types';
```
andreecosta, line 10: I think instead of creating only one file for all the contacts actions, it should be divided in a few files, one file for each action. (the same applies for guestUser and Titles) For example:
```
getContact.types
postContact.types
patchContact.types
...
```
and then, for example, on your **getContact.types** you should have this:

```
export type GetContact = (
  id: number,
  contactId: string,
  query?: Query,
  config?: Record<string, unknown>,
) => Promise<ContactResponse>;
```

and that **ContactResponse** should be, like you have, in contact.types.

So, one file for each action and use your contact.types for contact response (I think it should be the same for most of the actions).
You can check other examples like returns, bags, to get an idea of what I'm saying.
But if you need help you can talk to me :)

packages/client/src/users/types/contact.types.ts
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export type PatchContactData = {
+  value: string;
+  path: string;
+  op: string;
+  from: string;
+};
+
+export type PostContactQuery = {
```
andreecosta, line 29: I would put all query types in a single, separate file.
`query.types.ts`

packages/client/src/users/types/guestUser.types.ts
```diff
@@ -0,0 +1,57 @@
...
+export type SetPreferencesData = [
+  {
+    code: string;
+    values: string[];
+    groupId: string;
+    updatedDate: string;
+  },
+];
+
+export type PostGuestUser = (
```
andreecosta, line 37: There's a few more to move to individual files :eyes:

jpklzm: Fixed 👍🏻

andreecosta (approved):

packages/client/src/users/types/deleteContact.types.ts
```diff
@@ -0,0 +1,9 @@
+import type { ContactResponse } from './contact.types';
+import type { DeleteContactQuery } from './query.types';
+
+export type DeleteContact = (
+  id: number,
+  contactId: string,
+  query?: DeleteContactQuery,
+  config?: Record<string, unknown>,
```
joaoprcosta, line 8:
```suggestion
  import type { Config } from '../types';
  
  config?: Config,
```

packages/client/src/users/types/contact.types.ts
```diff
@@ -0,0 +1,20 @@
+export type CountryDetails = {
+  countrycode: string;
```
joaoprcosta, line 2: In other types, you use camelCase to define the name of the key. Could you please make it concistant?

packages/client/src/users/types/guestUser.types.ts
```diff
@@ -0,0 +1,10 @@
+export type GuestUserResponse = {
+  id: 0;
```
joaoprcosta, line 2:
```suggestion
  id: number;
```

packages/client/src/users/types/deleteContact.types.ts
jpklzm: fixed 👍🏻

packages/client/src/users/types/contact.types.ts
jpklzm: Nice catch! 👀  
Fixed 👍🏻

packages/client/src/users/types/guestUser.types.ts
jpklzm: fixed 👍🏻

pedro-gbf (approved):

PR #181, opened 2022-02-08
feat(react): move `handleAddOrUpdateItem` method to new hook

### LINKED ISSUE #16: Isolate `handleAddOrUpdateItem` from the `useBagItem` hook
## Is your proposal related to a problem?

Right now, the `useBagItem` hook states that the `bagItemId` argument is optional, but that is only true if we only want to use the `handleAddOrUpdateItem`. The whole hook besides this method is not ready to have the `bagItemId` argument as optional (Typescript helps a lot here).

```javascript
// Works correctly, but throws a typescript error - "Expected 1 argument, but got 0."
// If the argument was set as optional, the hook is very prone to error
const { handleAddOrUpdateItem } = useBagItem();
```

## Describe the solution you'd like

What would be nice here is to create another hook - `useAddOrUpdateItem` - isolating the content of the `handleAddOrUpdateItem` method, leaving the `useBagItem` hook with the correct mandatory argument. As for the current usage of the `useBagItem` hook, it has been needed only for that method, with no need for default values, hence another factor in favour of isolating. If, in fact, we have the need to use the method with the "default" bag item, we can use both hooks, like so

```javascript
const { bagItem } = useBagItem(myBagItemId);
const { handleAddOrUpdateItem } = useAddOrUpdateItem(bagItem);
```

anapim:
## Description

BREAKING CHANGE: This moves the method `handleAddOrUpdateItem` to a new hook `useAddOrUpdateBagItem`. The new hook is necessary since we are using selector and actions outside a React component.

To adapt this breaking change, we need to update all instances of:

```
const { handleAddOrUpdateItem } = useBagItem();
```

to

```
const handleAddOrUpdateItem = useAddOrUpdateBagItem();
```

Closes #16

It also changes the type of `customAttributes` to `unknown` because this is in fact an unknown value that depends on the implementation of each experience.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/bags/hooks/__tests__/useAddOrUpdateBagItem.test.tsx
+++ b/packages/react/src/bags/hooks/__tests__/useAddOrUpdateBagItem.test.tsx
@@ -1,0 +1,88 @@
+import { cleanup } from '@testing-library/react';
+import { getBagItems } from '@farfetch/blackout-redux/bags';
+import { mockBagItem, mockInitialState } from 'tests/__fixtures__/bags';
+import { mockStore } from '../../../../tests/helpers';
+import { Provider } from 'react-redux';
+import { renderHook } from '@testing-library/react-hooks';
+import { useAddOrUpdateBagItem } from '../..';
+import React from 'react';
+
+jest.mock('@farfetch/blackout-redux/bags', () => ({
+  ...jest.requireActual('@farfetch/blackout-redux/bags'),
+  addBagItem: jest.fn(() => ({ type: 'add' })),
+  updateBagItem: jest.fn(() => ({ type: 'update' })),
+  getBagItems: jest.fn(() => [mockBagItem]),
+}));
+
+const mockDispatch = jest.fn();
+
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
+  useDispatch: () => mockDispatch,
+}));
+
+const getRenderedHook = (state = mockInitialState) => {
+  const {
+    result: { current },
+  } = renderHook(() => useAddOrUpdateBagItem(mockBagItem), {
+    wrapper: props => <Provider store={mockStore(state)} {...props} />,
+  });
+
+  return current;
+};
+
+describe('useAddOrUpdateBagItem', () => {
+  beforeEach(jest.clearAllMocks);
+  afterEach(cleanup);
+
+  it('should return values correctly with initial state', () => {
+    const current = getRenderedHook();
+
+    expect(current).toStrictEqual(expect.any(Function));
+  });
+
+  describe('handleAddOrUpdateItem', () => {
+    it('should call handleAddOrUpdateItem and add item', async () => {
+      getBagItems.mockImplementationOnce(() => []);
+      const handleAddOrUpdateItem = getRenderedHook();
+
+      await handleAddOrUpdateItem(mockBagItem);
+
+      expect(mockDispatch).toHaveBeenCalledTimes(1);
+      expect(mockDispatch).toHaveBeenCalledWith({ type: 'add' });
+    });
+
+    it('should call handleAddOrUpdateItem and update item', async () => {
+      const handleAddOrUpdateItem = getRenderedHook();
+
+      await handleAddOrUpdateItem({ ...mockBagItem, quantity: 6 });
+
+      expect(mockDispatch).toHaveBeenCalledTimes(1);
+      expect(mockDispatch).toHaveBeenCalledWith({ type: 'update' });
+    });
+
+    it('should update and then add when current merchant has less stock than new quantity', async () => {
+      // Mock bag item has quantity 5
+      // The respective merchants have 7 and 5 available
+      const handleAddOrUpdateItem = getRenderedHook();
+
+      await handleAddOrUpdateItem({
+        quantity: 9,
+        size: {
+          ...mockBagItem.size,
+          stock: [
+            ...mockBagItem.size.stock,
+            {
+              merchantId: 213,
+              quantity: 5,
+            },
+          ],
+        },
+      });
+
+      expect(mockDispatch).toHaveBeenCalledTimes(2);
+      expect(mockDispatch).toHaveBeenCalledWith({ type: 'update' });
+      expect(mockDispatch).toHaveBeenCalledWith({ type: 'add' });
+    });
+  });
+});
--- a/packages/react/src/bags/hooks/useAddOrUpdateBagItem.ts
+++ b/packages/react/src/bags/hooks/useAddOrUpdateBagItem.ts
@@ -1,0 +1,149 @@
+/**
+ * Hook to provide all kinds of data for the business logic attached to the
+ * actions of adding or updating an item to the bag.
+ *
+ * @module useAddOrUpdateItem
+ * @category Bags
+ * @subcategory Hooks
+ */
+import {
+  addBagItem as addBagItemAction,
+  getBagItems,
+  updateBagItem as updateBagItemAction,
+} from '@farfetch/blackout-redux/bags';
+import {
+  buildBagItem,
+  generateBagItemHash,
+} from '@farfetch/blackout-redux/bags/utils';
+import { shallowEqual, useSelector } from 'react-redux';
+import { useAction } from '../../helpers';
+import type { HandleAddOrUpdateItem, UseAddOrUpdateBagItem } from './types';
+
+/**
+ * Provides handler for adding or updating a bag item.
+ *
+ * @memberof module:bags/hooks
+ *
+ * @param {number} [bagItem] - Bag item to work on.
+ *
+ * @returns {Function} Handle to manage bag item related operations.
+ */
+const useAddOrUpdateBagItem: UseAddOrUpdateBagItem = bagItem => {
+  // Selectors
+  const bagItems = useSelector(getBagItems, shallowEqual);
+  // Actions
+  const addBagItem = useAction(addBagItemAction);
+  const updateBagItem = useAction(updateBagItemAction);
+  const productSize = bagItem?.product?.sizes?.find(
+    size => size.id === bagItem?.size?.id,
+  );
+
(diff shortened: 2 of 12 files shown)


packages/react/src/bags/hooks/useBagItem.ts
```diff
@@ -431,13 +333,7 @@ const useBagItem: UseBagItem = bagItemId => {
      * @type {boolean}
      */
     isLoading,
-    /**
-     * @type {Function}
-     * @variation Member
-     *
-     * @see {@link module:useBagItem~handleAddOrUpdateItem|handleAddOrUpdateItem} method
-     */
-    handleAddOrUpdateItem,
```
mindouro, line 440: Wild idea: what if we still export `handleAddOrUpdateItem` with the bag item already provided, like `handleAddOrUpdateItem: handleAddOrUpdateItem(bagItem)`? Otherwise, the developer would probably end up importing both hooks 99% of the time it imports `useBagItem`, or am I missing something? 🤔

packages/react/src/bags/hooks/useBagItem.ts
```diff
@@ -210,8 +99,18 @@ const useBagItem: UseBagItem = bagItemId => {
     }
 
     const quantityToAdd = quantityToHandle - bagItem.quantity;
+    const productSize = bagItem?.product?.sizes?.find(
```
mindouro, line 102: If we provide the `bagItem` to the hook I think we can remove this logic from here 😁

packages/react/src/bags/hooks/useAddOrUpdateItem.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  return {
+    /**
+     * @type {Function}
+     * @variation Member
+     *
+     * @see {@link module:useAddOrUpdateItem~handleAddOrUpdateItem|handleAddOrUpdateItem} method
+     */
+    handleAddOrUpdateItem,
+  };
```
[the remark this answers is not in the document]
anapim: Done 🙂

packages/react/src/bags/hooks/useAddOrUpdateItem.ts
```diff
@@ -0,0 +1,147 @@
...
+   * item to handle.
+   * @param {object} [item.product=bagItem.product] - Product of the bag item
+   * to handle. Defaults to the product of the bag item that instantiated the
+   * hook.
+   * @param {number} [item.productAggregatorId = bagItem.productAggregator.id] - Product
+   * aggregator id that represents the bundle variant that owns it.
+   * @param {number} [item.quantity=bagItem.quantity] - Quantity of the
+   * product to add/update. Defaults to the quantity of the bag item that
+   * instantiated the hook.
+   * @param {string} [item.from] - Provenience of action.
```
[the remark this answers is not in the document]
anapim: It was but I think that it was not on purpose because it was declared in the types. I added it here 🙂

packages/react/src/bags/hooks/useAddOrUpdateItem.ts
```diff
@@ -0,0 +1,147 @@
...
+   * This is useful for when the operation is done outside the bag, for
+   * example, on the "move to bag" from the wishlist.
+   * <br />
+   * <br />
+   * <i><small>This operation is over the bag item that instantiated the hook by
+   * default.</small></i>.
+   *
+   * @function handleAddOrUpdateItem
+   * @param {object} [item] - Item to add/update the bag with.
+   * @param {string} [item.customAttributes=''] - Custom attributes of the bag
```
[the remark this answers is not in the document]
anapim: Done 🙂

packages/react/src/bags/hooks/useAddOrUpdateItem.ts
```diff
@@ -0,0 +1,147 @@
...
+   * item to handle.
+   * @param {object} [item.product=bagItem.product] - Product of the bag item
+   * to handle. Defaults to the product of the bag item that instantiated the
+   * hook.
+   * @param {number} [item.productAggregatorId = bagItem.productAggregator.id] - Product
+   * aggregator id that represents the bundle variant that owns it.
+   * @param {number} [item.quantity=bagItem.quantity] - Quantity of the
+   * product to add/update. Defaults to the quantity of the bag item that
+   * instantiated the hook.
```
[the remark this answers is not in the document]
anapim: Done 🙂

packages/react/src/bags/hooks/useBagItem.ts
```diff
@@ -46,130 +42,23 @@ import type { StoreState } from '@farfetch/blackout-redux/src/types';
  * to manage any bag item or product bag-related operation.
  */
 const useBagItem: UseBagItem = bagItemId => {
+  const { handleAddOrUpdateItem } = useAddOrUpdateItem();
```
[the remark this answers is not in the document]
anapim: Yep, nice catch. I moved it down.

tests/__fixtures__/bags/bagItem.fixtures.ts
```diff
@@ -35,7 +35,7 @@ export const mockBagItemEntity = {
 };
 
 export const mockIdenticalBagItem = {
-  customAttributes: { a: 1, b: 2, c: { d: 3 } },
```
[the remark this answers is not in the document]
anapim: I'm not sure, but according to the type for `customAttributes` we have
```
export type CustomAttributesAdapted = Record<string, string> | string | null;
``` 
which does not accept a number or an object as a value. Let me know if you think this is wrong.

packages/react/src/bags/hooks/useBagItem.ts
anapim: Removed!

packages/react/src/bags/hooks/useBagItem.ts
anapim: This approach would call `handleAddOrUpdateItem` twice, so it would need to be tweaked. 

Either way, I chose not to go forward with it, because in our use case scenarios we never have to call both hooks at the same time (mainly because, we will only use `useAddOrUpdateItem` to add and never to update, since `useBagItem` already handles the updates we need). The cases you are referring to happen only when we want to update the bag item and honestly I don't see a problem with that (in fact, it's exactly what is suggested in #16 )

packages/react/src/bags/hooks/useAddOrUpdateItem.ts
```diff
@@ -0,0 +1,147 @@
...
+   * stock, and so on.
+   * <br />
+   * This is useful for when the operation is done outside the bag, for
+   * example, on the "move to bag" from the wishlist.
+   * <br />
+   * <br />
+   * <i><small>This operation is over the bag item that instantiated the hook by
+   * default.</small></i>.
+   *
+   * @function handleAddOrUpdateItem
```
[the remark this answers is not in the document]
anapim: Removed.

packages/react/src/bags/hooks/useAddOrUpdateBagItem.ts
```diff
@@ -0,0 +1,148 @@
...
+        productAggregatorId,
+        quantity: quantityToAdd,
+        size,
+        ...otherParams,
+      });
+      // Checks if the item we want to add is already in bag
+      // by comparing the bag items' hash
+      const hash = generateBagItemHash(requestData);
+      const itemInBag =
+        bagItems && bagItems.find(item => generateBagItemHash(item) === hash);
```
nelsonleite, line 108:
```suggestion
        bagItems?.find(item => generateBagItemHash(item) === hash);
```

Small nitpick 😇

packages/react/src/bags/hooks/useAddOrUpdateBagItem.ts
```diff
@@ -0,0 +1,148 @@
+/**
+ * Hook to provide all kinds of data for the business logic attached to the
+ * actions of adding or updating a item to the bag.
```
nelsonleite, line 3:
```suggestion
 * actions of adding or updating an item to the bag.
```

packages/react/src/bags/hooks/__tests__/useAddOrUpdateBagItem.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+  beforeEach(jest.clearAllMocks);
+  afterEach(cleanup);
+
+  it('should return values correctly with initial state', () => {
+    const current = getRenderedHook();
+
+    expect(current).toStrictEqual(expect.any(Function));
+  });
+
+  describe('handleAddOrUpdateItem', () => {
```
nelsonleite, line 44: Do you think it's possible to add a test case for when the item has multiple merchants and it performs an update and an add operation? Since this is the main purpose of this hook, we could probably add a test ensuring exactly that 🙂

nelsonleite (changes requested):

packages/react/src/bags/hooks/useAddOrUpdateBagItem.ts
anapim: Done 🙂

packages/react/src/bags/hooks/useAddOrUpdateBagItem.ts
anapim: Done 🙂

packages/react/src/bags/hooks/__tests__/useAddOrUpdateBagItem.test.tsx
anapim: I have added a test, can you please check it out?

nelsonleite (approved):

tests/__fixtures__/bags/bagItem.fixtures.ts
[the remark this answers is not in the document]
anapim: After offline discussion, the type of `customAttributes` was changed to `unknown`.

mindouro (approved):

PR #199, opened 2022-02-18
feat(core): refactored mapper on omnitracking integration to calculate parameters from data object

ruifcnunes:
## Description
Refactored the integration and combined two different mappers into one, so we can calculate parameters based on the event data object in a single point.

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/analytics/integrations/Omnitracking/Omnitracking.js
+++ b/packages/core/src/analytics/integrations/Omnitracking/Omnitracking.js
@@ -26,11 +26,14 @@ import {
   OPTION_TRANSFORM_PAYLOAD,
 } from './constants';
 import { postTrackings } from './client';
-import { userGenderValuesMapper } from './definitions';
+import { trackEventsMapper, userGenderValuesMapper } from './definitions';
 import { v4 as uuidv4 } from 'uuid';
 import analyticsTrackTypes from '../../types/trackTypes';
 import get from 'lodash/get';
 import Integration from '../Integration';
+import isArray from 'lodash/isArray';
+import isFunction from 'lodash/isFunction';
+import isObject from 'lodash/isObject';
 import logger from '../../utils/logger';
 import platformTypes from '../../types/platformTypes';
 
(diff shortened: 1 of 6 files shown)


danielbento92 (approved):

packages/core/src/analytics/integrations/__tests__/Omnitracking.test.js
```diff
@@ -352,9 +353,10 @@ describe('Omnitracking', () => {
     });
     describe('Should send track events when', () => {
       it('event is `Place Order Started`', async () => {
-        const tid = 10097;
+        const tid1 = 10097;
```
boliveira, line 356: Nitpicking: Is it possible to give more descriptive names for those tids? It would be easier if they have the corresponding action name to identify what they mean. If it is not possible, you can disregard this comment.

packages/core/src/analytics/integrations/Omnitracking/Omnitracking.js
```diff
@@ -215,26 +218,56 @@ class Omnitracking extends Integration {
...
-          formatTrackEvent(data, precalculatedParameters),
-          data,
-        );
+        if (eventMapperFn) {
+          const mappedEvents = eventMapperFn(data);
+
+          if (isArray(mappedEvents)) {
+            await Promise.all(
+              mappedEvents.map(async mappedEventData => {
+                await this.processTrackEvents(data, mappedEventData);
```
boliveira, line 235: I think this await might be unnecessary, because it will prevent the requests to be executed in parallel. Is there any problem in removing the `await` from this call and only returning the result of `this.processTrackEvents(data, mappedEventData)`? Without the `await` those events would arrive to the server out-of-order, so if it is necessary that they arrive in order, keep the `await` instruction.

packages/core/src/analytics/integrations/Omnitracking/Omnitracking.js
```diff
@@ -215,26 +218,56 @@ class Omnitracking extends Integration {
...
-        precalculatedParameters = this.getPrecalculatedParametersForEvent(data);
+        const eventMapperFn = trackEventsMapper[data.event];
 
-        if (!this.validateTrackingRequisites()) {
-          logger.warn(
-            `[Omnitracking] - Event ${data.event} tried to track without an unique view id.
-                Consider tracking an event only after a page event.
-                This behaviour is not recommended and will not be supported in upcoming versions`,
+        if (eventMapperFn && !isFunction(eventMapperFn)) {
+          throw new Error(
```
boliveira, line 224: This error seems to imply that the user is making a configuration mistake, however the `trackEventsMapper` is internal. Maybe the best here would be to have a unit test that checked if all values from all keys of `trackEventsMapper` are a function?

packages/core/src/analytics/integrations/Omnitracking/Omnitracking.js
```diff
@@ -215,26 +218,56 @@ class Omnitracking extends Integration {
...
+          if (isArray(mappedEvents)) {
+            await Promise.all(
+              mappedEvents.map(async mappedEventData => {
+                await this.processTrackEvents(data, mappedEventData);
+              }),
+            );
+          }
+
+          if (isObject(mappedEvents)) {
+            return await this.processTrackEvents(data, mappedEvents);
```
boliveira, line 241: I think the await is unnecessary here because the return value is not being used by `analytics` when invoking the `track` function of all integrations.

boliveira (approved):

packages/core/src/analytics/integrations/__tests__/Omnitracking.test.js
ruifcnunes: Fixed on dev-main-22-02-22. thanks 🙂

packages/core/src/analytics/integrations/Omnitracking/Omnitracking.js
ruifcnunes: Fixed on dev-main-22-02-22. thanks 🙂

packages/core/src/analytics/integrations/Omnitracking/Omnitracking.js
ruifcnunes: Fixed on dev-main-22-02-22. thanks 🙂

packages/core/src/analytics/integrations/Omnitracking/Omnitracking.js
ruifcnunes: Fixed on dev-main-22-02-22. thanks 🙂

PR #202, opened 2022-02-22
feat(core): add new mapper on omnitracking integration to calculate parameters from data object - next

ruifcnunes:
## Description
Refactored the integration and combined two different mappers into one, so we can calculate parameters based on the event data object in a single point.

## Checklist

## Disclaimer

### DIFF
--- a/packages/analytics/src/integrations/Omnitracking/Omnitracking.js
+++ b/packages/analytics/src/integrations/Omnitracking/Omnitracking.js
@@ -26,11 +26,13 @@ import {
   OPTION_TRANSFORM_PAYLOAD,
 } from './constants';
 import { postTrackings } from '@farfetch/blackout-client/omnitracking';
-import { userGenderValuesMapper } from './definitions';
+import { trackEventsMapper, userGenderValuesMapper } from './definitions';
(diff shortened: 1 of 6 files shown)


packages/analytics/src/integrations/Omnitracking/Omnitracking.js
```diff
@@ -215,29 +217,61 @@ class Omnitracking extends Integration {
...
-          );
-
-          return;
+        const eventMapperFn = trackEventsMapper[data.event];
+
+        if (eventMapperFn) {
+          const mappedEvents = eventMapperFn(data);
+
+          if (isArray(mappedEvents)) {
+            await Promise.all(
```
boliveira, line 226: I think the `await` is not needed because this function is called by analytics which does not await the result, so it is safe to return the promise or just have a return statement after the Promise.all call.

boliveira (approved):

packages/analytics/src/integrations/Omnitracking/Omnitracking.js
ruifcnunes: fixed, thanks :)

danielbento92 (approved):

Issue #107, opened 2022-01-06
Replace jsdoc with tsdoc


boliveira:
**Motivation**

As we are migrating the code to typescript, we need to change our usage of jsdoc to match the tsdoc specification, which does not require that we declare the types of parameters / return values of functions as they are already available on the typescript type declarations.

To make this work, we will need to replace our eslint-plugin-jsdoc with eslint-plugin-tsdoc. 

You can get more information by checking the [official tsdoc documentation page](https://tsdoc.org/).

**jsdoc & better-docs**

We use better-docs to generate our documentation from the jsdoc. When migrating to tsdoc, the documentation might break. As it is not being used right now, it is not a major concern. However, there are some jsdoc tags that become unnecessary in tsdoc as they were specific for better-docs to generate some special output. By using the `eslint-plugin-tsdoc` plugin, these tags will give errors when linting, so we will need to remove them as well.

**Dependencies**

This issue should be tackled during (or after) the migration to typescript.

PR #209, opened 2022-02-28
feat(analytics|react): convert analytics area from js to ts

### LINKED ISSUE #15: Remove (or refactor) `createInstance` method from analytics's `Integration` class
## Is your proposal related to a problem?

Currently, to create a new integration in the analytics package, it is necessary that the developer overrides the `Integration` class's static method `createInstance` beyond creating a constructor. At the time, this method was created to provide more flexibility for the `Integration` on the object it can return. For example, an integration could easily use the Decorator pattern to return a wrapper to the real instance that does additional logging or preprocessing of the event before it is received by the final instance that would process the event.

While this feature is a good thing to have, in practice that necessity never materialised and having to force all integrations to override this method for the integration to work, even if they only end up returning a naked instance of the same class, ends up creating unnecessary additional work for the developer.

## Describe the solution you'd like

Either remove this method from the `Integration` API and have the analytics instantiate the `Integration`'s subclass directly or provide a way to keep this feature without requiring to override this method if the developer just wants to return a simple instance of the class.

danielbento92:
## Description

- Convert analytics core and react area to typescript.
- This also replaces `jsdoc` with `tsdoc`. The `docs`
script (and all dependencies used by it) were removed
as they would not work with `tsdoc`. This begins the work
for the issue [#107](https://github.com/Farfetch/blackout/issues/107)
but is not complete as all teams will need to review and
fix any annotations that are not correct in tsdoc.
- Some introductory work was done for using project
references in typescript in order to support the `analytics`
package compilation that imports `package.json` which
would not work with a single ts configuration file because
the rootDir was `src` and `package.json` was outside it.
Using project references in typescript allows to have the
`package.json` included alongside the src folder contents 
while also keeping the layout for the compiled package
to not include the `src` folder.

Refs #107
Closes #15 

### Dependencies

None.

## Checklist

## Disclaimer


packages/analytics/src/integrations/Integration.ts
```diff
@@ -0,0 +1,105 @@
+/* eslint-disable no-unused-vars */
+
+import type ConsentData from '../types/consentData.types';
+
+export interface IntegrationFactory {
+  new (
+    options: Record<string, unknown>,
+    loadData: Record<string, unknown>,
```
boliveira, line 8: loadData should be typed as the `EventData` type you have in `eventData.types`. However, there is one nuance here: The `type` and `event` properties contain the value `loadIntegration`. You might need to define additional types for those event cases, for example you might define a LoadIntegrationEventType that extends EventData and override the `type` and `event` properties with the type `loadIntegration` and use that type to type the loadData parameter..

packages/analytics/src/integrations/Integration.ts
```diff
@@ -0,0 +1,105 @@
...
+    strippedDownAnalytics: Record<string, unknown>,
+  ): Integration;
+  createInstance(
+    options: Record<string, unknown>,
+    loadData: Record<string, unknown>,
+    analytics: Record<string, unknown>,
+  ): Integration;
+  shouldLoad(consent: Record<string, unknown>): boolean;
```
boliveira, line 16: `consent` parameter should be typed with `ConsentData` type.

packages/analytics/src/Analytics.ts
```diff
@@ -142,10 +179,10 @@ class Analytics {
   /**
    * Sets the consent and passes it to the registered integrations.
    *
-   * @param {object} data - Consent object to be stored.
-   * @returns {Promise<Analytics>} Promise that will resolve with the instance that was used when calling this method to allow chaining.
+   * @param data - Consent object to be stored.
```
ruifcnunes, line 182: Nitpick: add an empty line between the @params and the @return. valid for all cases in this document

packages/analytics/src/Analytics.ts
```diff
@@ -535,71 +615,82 @@ class Analytics {
...
    *
-   * @returns {Promise<object>} - User data to be sent to integrations.
-   * @private
+   * @returns User data to be sent to integrations.
    */
-  async getSetUserEventData() {
-    return await this.getEventData('onSetUser', {
-      event: 'onSetUser',
+  protected async getSetUserEventData(): Promise<SetUserEventData> {
+    const val = await this.getEventData(ON_SET_USER_TRACK_TYPE, {
```
ruifcnunes, line 659: do we need this extra val const?

packages/analytics/src/Consent.ts
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @returns Promise that will resolve with the consent data.
+   */
+  async get(): Promise<ConsentData> {
+    const getItem = (await this.storage.getItem(
+      'consent',
+    )) as Partial<ConsentData>;
+
+    return {
+      marketing: getItem?.marketing || defaultConsentData.marketing,
```
ruifcnunes, line 48: Here we cannot return the default values if nothing returns from the storage. The functionality is:
If there is no consent given, this method must return `null`. If the user interacted with the consent (allowing these categories or not), we return what was previously stored.
The return `null` case allows tenants to understand if the user did interact with the consent already or not (e.g.: "Accept all cookies" banner, otherwise we wouldn't have a way to understand if the user didn't give permissions at all (that would lead to have all categories with "false" value, which is the default as well).

packages/analytics/src/Consent.ts
```diff
@@ -0,0 +1,79 @@
...
+   * @returns Promise that will resolve with the instance that was used when calling this method to allow chaining.
+   */
+  async set(data: Partial<ConsentData> = defaultConsentData): Promise<Consent> {
+    const consent = pick(data, CONSENT_KEYS);
+    const currentConsentFromStorage = ((await this.storage.getItem(
+      'consent',
+    )) || {}) as typeof data;
+
+    const newConsent = merge(currentConsentFromStorage, {
+      marketing: consent.marketing || !!currentConsentFromStorage?.marketing,
```
ruifcnunes, line 67: Couldn't we directly merge the previous consent with the new one without having these extra validations, like we had before?

packages/analytics/src/DataStore.ts
```diff
@@ -0,0 +1,43 @@
+import merge from 'lodash/merge';
+/**
+ * This is a simple helper class to hold data in memory and encapsulates get/set data logic.
+ */
+class DataStore {
+  data: Record<string, unknown>;
```
ruifcnunes, line 6: Could we store this into a type and reference it wherever it is used (get and set methods)?

packages/analytics/src/types/analytics.types.ts
```diff
@@ -0,0 +1,69 @@
...
+export type EventContext = ContextData & { event: Record<string, unknown> };
+
+export type EventData<T> = {
+  type: T;
+  event: T extends typeof LOAD_INTEGRATION_TRACK_TYPE
+    ? typeof LOAD_INTEGRATION_TRACK_TYPE
+    : T extends typeof ON_SET_USER_TRACK_TYPE
+    ? typeof ON_SET_USER_TRACK_TYPE
+    : string;
+  user: unknown;
```
ruifcnunes, line 32: Can we use `UserData` type here?

ruifcnunes (commented):
Left some comments, can you please check them?

packages/analytics/src/Analytics.ts
boliveira: Done! Thanks!

packages/analytics/src/Analytics.ts
boliveira: No, we do not, it was a leftover of a test I was doing and I forgot to rollback. Reverted, thanks!

packages/analytics/src/Consent.ts
boliveira: Yes, my bad, haven't checked against the previous implementation. Reverted, thanks!

packages/analytics/src/Consent.ts
boliveira: Yes we could, changed ConsentData to be a Partial by default, so we can do it. Thanks!

packages/analytics/src/DataStore.ts
boliveira: Yes, created a `Data` type for that. Thanks!

packages/analytics/src/types/analytics.types.ts
boliveira: Yes, forgot to change that case. In this case it has to be `UserData | null` as the `user` property is populated by the call to `this.user()` method which can return null. Thanks!

boliveira (approved):

PR #239, opened 2022-03-25
fix(client): unable to manage wishlist of an oos product

ricardojmf:
## Description

This PR solves the problem regarding a product that is added to the
wishlist with a size and then becomes OOS.
In this situation the API returns a wishlistItem with one size attribute
instead of an empty array so this solution passes by reproducing the same
login for one attribute array as for an empty attribute array.

### Dependencies

None.

## Checklist

## Disclaimer


ricardojmf:
Another solution found.

PR #255, opened 2022-04-07
feat(react): add castle 2.0 analytics' integration - Next

ruifcnunes:
## Description

This commit adds the new version of the Castle integration using the Castle SDK 2.0.
**BREAKING CHANGE**:
From now on, the `appId` option is no longer valid.
Please be sure to pass the new `pk` option (castle publishable key), along with [any optional parameter you might need](https://docs.castle.io/docs/sdk-browser#configuration).

## Checklist

## Disclaimer


danielbento92 (approved):

packages/react/src/analytics/integrations/Castle/__tests__/Castle.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    it('Should return a Castle instance from createInstance', () => {
+      instance = createInstance();
+
+      expect(instance).toBeInstanceOf(Castle);
+    });
+  });
+
+  describe('Validation', () => {
+    it('Should log an error if no publishable key is passed via `options`', () => {
+      // @ts-ignore - We want to test this specific case
```
boliveira, line 73:
```suggestion
      // @ts-expect-error - We want to test this specific case
```
Using // @ts-expect-error here is better because it makes explicit that you are expecting the construct on the next line to be wrong and if in the future that construct is not an error, for example `createInstance` accepts a `null` value for `pk` property, you will get an error in typescript, which will help you to either remove or change the unit test.

packages/react/src/analytics/integrations/Castle/__tests__/Castle.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      expect(() => createInstance({ pk: null })).toThrowError();
+    });
+
+    it('Should throw an error if an invalid `configureHttpClient` is passed', () => {
+      expect(() =>
+        // @ts-ignore - We want to test this specific case
```
boliveira, line 79: Same as here https://github.com/Farfetch/blackout/pull/255/files#r845987247
There are other instances in this file as well which this comment is valid.

packages/react/src/analytics/integrations/Castle/types/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import type {
+  configure,
+  createRequestToken,
+  custom,
+  form,
+  page,
+} from '@castleio/castle-js';
+import type { IntegrationOptions } from '@farfetch/blackout-analytics';
+
+export interface CastleIntegrationOptions extends IntegrationOptions {
```
boliveira, line 11: Since `CastleIntegrationOptions` is baed on the ConfigureOptions from `castle` package, this could be done instead:

```suggestion

import type {
  ConfigureOptions,
} from '@castleio/castle-js';

export interface CastleIntegrationOptions extends IntegrationOptions {
configureOptions: ConfigureOptions
debugModeOn?: boolean;
  httpClient?: AxiosInstance;
  clientIdHeaderName?: string;
  configureHttpClient?: (
    tokenCreator: typeof createRequestToken,
  ) => Promise<void>;
```

If a separate property is used for the configure options, if there is a new version of '@castleio/castle-js' package that adds new properties, the integration code will not need to be changed. However for this to work, you will need to change this call as well:

```js
applyOptions(options: CastleIntegrationOptions) {
    try {
      this.castleJS.configure(options.configureOptions); <--- Just pass the configureOptions prop directly into the method

      this.installInterceptor(options);
    } catch (error) {
      throw new Error(
        `${CASTLE_MESSAGE_PREFIX} Failed to initialize Castle. Error: ${error}`,
      );
    }
```
What do you think?

packages/react/src/analytics/integrations/Castle/Castle.ts
```diff
@@ -0,0 +1,328 @@
...
+export const CASTLE_MESSAGE_PREFIX = 'Castle 2.x -';
+
+/**
+ * Castle integration.
+ */
+class Castle extends Integration<CastleIntegrationOptions> {
+  httpClientInterceptor?: number;
+  isInterceptorAttached?: boolean;
+  debugModeOn: boolean;
+  castleJS: CastleInstance;
```
boliveira, line 56: The `CastleInstance` type can be removed and replaced by this which is more type-safe:

```js
castleJS: typeof castleJS
```

packages/react/src/analytics/integrations/Castle/Castle.ts
```diff
@@ -0,0 +1,328 @@
...
+   */
+  static shouldLoad() {
+    return true;
+  }
+
+  /**
+   * Creates an instance of Castle integration.
+   *
+   * @param options - Integration options.
+   * @param loadData - Analytics's load event data.
```
boliveira, line 73: Missing `analytics` parameter in jsdoc

boliveira (approved):

Issue #114, opened 2022-01-10
Redux actions error handling


boliveira:
## Is your proposal related to a problem?

The redux actions are not checking if the error that is catched is effectively 
an Error object and is blindly dispatching the catched error to the store, 
which might have an impact in the consumer's code if they try to access
properties of an Error object on a value that is not an Error.

For example:

```javascript

// Example from fetchFormSchemaFactory.js

try {
      const result = await getFormSchema(schemaCode, query, config);

      dispatch({
        meta: { schemaCode },
        payload: result,
        type: actionTypes.FETCH_FORM_SCHEMA_SUCCESS,
      });

      return result;
    } catch (error) {

      dispatch({
        meta: { schemaCode },
        payload: { error }, // We blindly dispatch the error here without checking. The error property is typed in typescript as receiving an Error object.
        type: actionTypes.FETCH_FORM_SCHEMA_FAILURE,
      });

      throw error;
    }

```

## Describe the solution you'd like

We need to add code to all actions that guarantees that an Error is always what will be
available on the store for the consumers. A helper function could be defined that
would return a proper Error object from the error that is catched from the action.

Example:

```javascript

try {
      const result = await getFormSchema(schemaCode, query, config);

      dispatch({
        meta: { schemaCode },
        payload: result,
        type: actionTypes.FETCH_FORM_SCHEMA_SUCCESS,
      });

      return result;
    } catch (error) {

      // The wrapError function would return always an Error object.
      // If `error` is already an Error object, it would just return it.
      // Else, it would wrap the error in an Error object.
      const finalError = helpers.wrapError(error);

      dispatch({
        meta: { schemaCode },
        payload: { error: finalError },
        type: actionTypes.FETCH_FORM_SCHEMA_FAILURE,
      });

      throw error;
    }

```

The wrapError function could return a subclass from Error, like UnknownError when the `error` argument is not
an `Error` object. It could contain a property `originalError` which contained the error argument 
and its `message` property could be formatted something like this: ``Unknown error: ${originalError?.toString()}``

## Additional information

This issue should be tackled before migration to typescript 4.4 if we want to use the compiler option [`--useUnknownInCatchVariables`](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables) introduced by it which will default the type of catch variables to `unknown` instead of `any`.

PR #307, opened 2022-05-17
feat(blackout-react): fix issues for the release blackout 2

danielbento92:
## Description

- move mocks to fixtures folders on omnitracking and recentlyViewed.
- add cspell configuration.
- fix analytics package warnings.
- fix many analytics integrations tsdoc warnings.
- fix omnitracking missing parameters.
- fix some wrong words in analytics context.
- perform reset from store, of recommendations and recently viewed
when logout action is called.

### Dependencies

None.

## Checklist

## Disclaimer


packages/redux/src/recentlyViewed/types/action.types.ts
```diff
@@ -12,7 +12,7 @@ interface FetchRecentlyViewedProductsSuccessAction extends Action {
 }
 interface FetchRecentlyViewedProductsFailureAction extends Action {
   type: typeof actionTypes.FETCH_RECENTLY_VIEWED_PRODUCTS_FAILURE;
-  payload: { error: Error };
+  payload: { error: Error | unknown };
```
boliveira, line 15: `Error | unknown` is effectively `unknown` as you can see here: https://github.com/microsoft/TypeScript/blob/b975dfa1027d1f3073fa7cbe6f7045bf4c882785/tests/baselines/reference/unknownType1.types

If you want to remove a warning, then it should be typed as `unknown`. However, that might need other code changes related with this issue https://github.com/Farfetch/blackout/issues/114

packages/client/src/forms/__tests__/getFormSchema.test.ts
```diff
@@ -32,6 +32,7 @@ describe('schemas client', () => {
       ).resolves.toEqual(successResponse);
 
       expect(spy).toHaveBeenCalledWith(
+        // @ts-expect-error
```
boliveira, line 35: This `// @ts-expect-error` should not be added as the real problem is with the type `FormSchemaQuery` that should be defined like this:

```js
export type FormSchemaQuery = {
  includeJsonSchema?: 'true' | 'false';
  resolveJsonSchemaPresets?: 'true' | 'false';
};
```

This is because of the typings used for the `proper-url-join` which do not accept boolean values for the `query` property. Can you change that type and remove any unnecessary `// @ts-expect-error` directives that might result of that change?

output.txt
```diff
@@ -0,0 +1,10 @@
+packages/redux/src/forms/actions/factories/fetchFormSchemaFactory.ts(39,20): error TS2322: Type 'unknown' is not assignable to type 'Error'.
```
ruifcnunes, line 1: should this file be here?

packages/client/src/omnitracking/postBatchTrackings.ts
```diff
@@ -24,7 +24,10 @@ import type { AxiosRequestConfig } from 'axios';
  *
  * @returns Promise object.
  */
-const postBatchTrackings = (data: Array<object>, config?: AxiosRequestConfig) =>
+const postBatchTrackings = (
+  data: Array<Record<string, unknown>>,
```
boliveira, line 28: There is an inconsistency between this type and the type defined for `data` in `postTrackings` client. The type for `data` in `postBatchTrackings` should be an array of elements of the same type defined for `data` in the `postTrackings` client. 

An alias should be defined to help with that, like this:

```js
type TrackingData = Record<string, unknown>;

// postBatchTrackings.ts
const postBatchTrackings = (
  data: Array<TrackingData>,
...

// postTrackings.ts
const postTrackings = (data: TrackingData, config?: AxiosRequestConfig) =>
...
```

output.txt
boliveira, line 1: This can be removed

boliveira (approved):

ruifcnunes (approved):

output.txt
danielbento92: Fixed!

packages/client/src/omnitracking/postBatchTrackings.ts
danielbento92: Fixed!

output.txt
danielbento92: Fixed!

packages/client/src/forms/__tests__/getFormSchema.test.ts
boliveira: As discussed offline, we will keep as is.

PR #322, opened 2022-05-31
fix(redux): get entity selectors errors

andreecosta:
## Description

- Fix getEntityById selector
- Fix getEntityById implementation erros

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/bags/selectors.ts
+++ b/packages/redux/src/bags/selectors.ts
@@ -104,16 +104,20 @@ export const getBagId = (state: StoreState): State['id'] => getId(state.bag);
  *     bagItem: getBagItem(state, id)
  * });
  */
-
+// @TODO Remove cast from functions
 export const getBagItem = createSelector(
   [
-    (state: StoreState, bagItemId: BagItem['id']): BagItemEntity =>
-      getEntityById(state, 'bagItems', bagItemId),
+    (state: StoreState, bagItemId: BagItem['id']) =>
+      getEntityById(state, 'bagItems', bagItemId) as BagItemEntity,
     (
       state: StoreState,
       bagItemId: BagItem['id'],
     ): ProductEntity | undefined => {
-      const bagItem = getEntityById(state, 'bagItems', bagItemId);
+      const bagItem = getEntityById(
+        state,
+        'bagItems',
+        bagItemId,
+      ) as BagItemEntity;
 
       return getProduct(state, bagItem?.product);
     },
--- a/packages/redux/src/checkout/selectors.ts
+++ b/packages/redux/src/checkout/selectors.ts
@@ -80,7 +80,7 @@ export const getCheckoutId = (state: StoreState): State['id'] =>
 export const getCheckout = (state: StoreState): CheckoutEntity | undefined => {
   const checkoutId = getCheckoutId(state);
 
-  return checkoutId && getEntityById(state, 'checkout', checkoutId);
+  return checkoutId ? getEntityById(state, 'checkout', checkoutId) : undefined;
 };
 
 /**
@@ -109,7 +109,9 @@ export const getCheckoutOrder = (
 ): CheckoutOrderEntity | undefined => {
   const checkoutId = getCheckoutId(state);
 
-  return checkoutId && getEntityById(state, 'checkoutOrders', checkoutId);
+  return checkoutId
+    ? getEntityById(state, 'checkoutOrders', checkoutId)
+    : undefined;
 };
 
 /**
@@ -126,7 +128,9 @@ export const getCheckoutDetail = (
 ): CheckoutDetailsEntity | undefined => {
   const checkoutId = getCheckoutId(state);
 
-  return checkoutId && getEntityById(state, 'checkoutDetails', checkoutId);
+  return checkoutId
+    ? getEntityById(state, 'checkoutDetails', checkoutId)
+    : undefined;
 };
 
 /**
--- a/packages/redux/src/entities/selectors/entity.ts
+++ b/packages/redux/src/entities/selectors/entity.ts
@@ -1,8 +1,9 @@
-import get from 'lodash/get';
 import invariant from 'invariant';
 import type { StoreState } from '../../types';
 
 type StateEntities = StoreState['entities'];
+type SchemaName = keyof NonNullable<StateEntities>;
+type EntityId = keyof NonNullable<NonNullable<StateEntities>[SchemaName]>;
 
 const throwError = (state: StoreState) => {
   invariant(
@@ -17,17 +18,13 @@ const throwError = (state: StoreState) => {
 /**
  * Gets all entities of a given schema name.
  *
- * @memberof module:entities/selectors
+ * @param state - Application state.
+ * @param name - Entity/schema name.
  *
- * @param {object} state - Application state.
- * @param {string} name - Entity/schema name.
- *
- * @returns {object|undefined} - All entities of the given schema name, undefined
+ * @returns - All entities of the given schema name, undefined
  * if none is found.
  */
-export const getEntities = <
-  SchemaName extends keyof NonNullable<StateEntities>,
->(
+export const getEntities = (
   state: StoreState,
   name: SchemaName,
 ): NonNullable<StateEntities>[SchemaName] | never => {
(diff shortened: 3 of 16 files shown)


pedro-gbf (approved):

packages/redux/src/entities/selectors/entity.ts
```diff
@@ -17,17 +19,13 @@ const throwError = (state: StoreState) => {
 /**
  * Gets all entities of a given schema name.
  *
- * @memberof module:entities/selectors
+ * @param state - Application state.
```
catarinacouto, line 22: I believe here we can leave the types as before: `* @param {string} name`. 

What do you think? If fits, apply this to every place that makes sense.

andreecosta: I can undo, but it has a warning since this is a TS file.
Since the function is correctly typed the tsDoc correctly types those parameters too.

jpklzm (approved):

packages/redux/src/entities/selectors/entity.ts
boliveira: Hello! As the project is now configured with `tsdoc`, we need to remove all `jsdoc` typing information as it is not necessary for `tsdoc`. We have a task to remove them all from the package.

packages/redux/src/entities/selectors/entity.ts
```diff
@@ -42,36 +40,27 @@ export const getEntities = <
...
-  SchemaName extends keyof NonNullable<StateEntities>,
-  EntityId extends keyof NonNullable<NonNullable<StateEntities>[SchemaName]>,
->(
+export const getEntityById = <T>(
   state: StoreState,
   name: SchemaName,
   id: EntityId,
-): NonNullable<StateEntities>[SchemaName][EntityId] | undefined =>
-  getEntities(state, name)?.[id];
+): T => getEntities(state, name)?.[id];
```
nelsonleite, line 54:
```suggestion
): T | undefined => getEntities(state, name)?.[id];
```

Not sure about this but, what about having the `undefined` part explicit here (because it's a reality since we have optional chaining) instead of having the caller to always add `| undefined`? Mainly because the caller can send what it wants, and if the undefined is missing, the types won't be correct.

For example here

```js
export const getAddress = (args): RandomType => getEntityById(args);
```

If I just set `RandomType`, it won't be true, because the result can actually be undefined as well... Does this make sense? 🤔

joaoprcosta (approved):

packages/redux/src/entities/selectors/entity.ts
```diff
@@ -42,36 +40,27 @@ export const getEntities = <
...
- *
- * @param   {object} state - Aplication state.
- * @param   {string} name  - Entity name.
- * @param   {string|number} id    - Entity identifier.
+ * @param state - Aplication state.
+ * @param name  - Entity name.
+ * @param id    - Entity identifier.
  *
- * @returns {object} The entity.
+ * @returns The entity.
```
nelsonleite, line 63: Probably out of scope but, having a `getEntities` and a `getEntityById`, do you think that our `getEntity` still makes sense?

If I recall correctly, `getEntity` was the previous selector we used, and it behaves somewhat unexpected. I'm just asking this because the documentation and the goal is very similar to what the other selectors do, but with no side effects. Food for thought 🙂

packages/redux/src/returns/selectors.ts
```diff
@@ -68,8 +69,10 @@ export const getReturnError = (state: StoreState) => getError(state.returns);
  *
  * @returns {object} Return item.
  */
-export const getReturnItem = (state: StoreState, returnItemId: string) =>
-  getEntity(state, 'returnItems', returnItemId);
+export const getReturnItem = (
+  state: StoreState,
+  returnItemId: string,
+): ReturnItem => getEntity(state, 'returnItems', returnItemId);
```
nelsonleite, line 75: My previous comment is reflected here - why don't we use `getEntityById` instead?

nelsonleite (changes requested):

packages/redux/src/entities/selectors/entity.ts
catarinacouto: Ok, in these scenarios we can remove the jsdoc types :) But in this PR it's not applied the delete of these types in every file we update. Can we have this?

packages/redux/src/checkout/selectors.ts
```diff
@@ -80,7 +80,7 @@ export const getCheckoutId = (state: StoreState): State['id'] =>
 export const getCheckout = (state: StoreState): CheckoutEntity | undefined => {
   const checkoutId = getCheckoutId(state);
 
-  return checkoutId && getEntityById(state, 'checkout', checkoutId);
+  return checkoutId ? getEntityById(state, 'checkout', checkoutId) : undefined;
```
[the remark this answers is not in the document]
andreecosta: If I just use leave the selector I get an error on checkoutId because getCheckoutId either returns a number or null, and getEntityById function expects a number or string.
Maybe that's why there was a check for checkoutId before the selector, but even that had to be changed since it returned more types than expected by the function.

Since this was not my area I didn't want to mess with the rest of the functions.

packages/redux/src/entities/selectors/entity.ts
andreecosta: I just removed because I was changing the function and I thought "why not". But I can undo it if you prefer to do it in a PR related to that task. Just let me know :)

packages/redux/src/entities/selectors/entity.ts
andreecosta: If I explicitly set undefined as a possible response in the function, I have to explicitly set the same whenever I call the function. Because when I'm calling it, and I say that I'm only expecting lets say `RandomType` as a response, typescript gives an error saying that the function may also return undefined.

packages/redux/src/entities/selectors/entity.ts
andreecosta: I've checked and there's still a few files using `getEntity`, although the descriprtion for both functions looks the same, I noticed that `getEntity` accepts a null id. And some of the implementations are not sending the id.

But, since all those implementations are done on my area (account), I can check with my team and see if we can change that and remove that function :)

packages/redux/src/returns/selectors.ts
andreecosta: Answered on previous comment. Checking with my team :)

packages/redux/src/entities/selectors/entity.ts
andreecosta: Ok, just checked. So, what the function does is return the entity we want, without an id, and we (account) have a few use cases for that. This is one of them for example:

```export const getMerchants = (state: StoreState) => getEntity(state, 'merchants');```

For those cases like the one you mentioned in the below comment, the one where we are passing an Id, I will change it to use the "new" function. But for the rest of the cases I have to keep using `getEntity`.

**Review by catarinacouto — APPROVED** [review]

File: packages/redux/src/entities/selectors/entity.ts (hunk 2 above)
Comment (nelsonleite) [reply packages/redux/src/entities/selectors/entity.ts:54]: And that's what's expected, right? I mean, if you just expect `RandomType` as a response, it's wrong, because it can actually be undefined as well, so the TS error makes sense, doesn't it? 🤔 _(If I understood correctly)_

File: packages/redux/src/entities/selectors/entity.ts (hunk 3 above)
Comment (nelsonleite) [reply packages/redux/src/entities/selectors/entity.ts:63]: Ok, we can let it be for now, but I strongly recommend opening an issue with this, because it makes no sense to maintain 3 functions, where one of them is a mix of the other 2 (those 2 who were born to solve the problem that the 3rd has) 🧐 

I don't remember where we had it reported, but in a nutshell, a `getEntity` might return an entity (which is correct) or return a bunch of entities, if the id of the one provided isn't found (which is a side-effect, actually, not the intended) - for example, for `getEntity(state, 'merchants', 123)`, if the merchant `123` doesn't exist, it'll return all merchant entities.

File: packages/redux/src/entities/selectors/entity.ts (hunk 2 above)
Comment (andreecosta) [reply packages/redux/src/entities/selectors/entity.ts]: done 😁

File: packages/redux/src/bags/selectors.ts [hunk 6]
```diff
@@ -107,13 +107,17 @@ export const getBagId = (state: StoreState): State['id'] => getId(state.bag);
...
     (
       state: StoreState,
       bagItemId: BagItem['id'],
     ): ProductEntity | undefined => {
-      const bagItem = getEntityById(state, 'bagItems', bagItemId);
+      const bagItem = getEntityById(
+        state,
+        'bagItems',
+        bagItemId,
+      ) as BagItemEntity;
```
mindouro, line 120: On "packages/redux/src/wishlists/selectors/wishlistsSets.ts" you removed the cast to `WishlistSetEntity` in favor of typing the const like `const wishlistSet: WishlistSetEntity | undefined = ...`. Can't we do the same here?

mindouro (changes requested):

packages/redux/src/entities/selectors/entity.ts
andreecosta: Yes, if it is a quick change I'll update on this PR, otherwise i'll create a task to be done in our team.

packages/redux/src/bags/selectors.ts
andreecosta: I tried doing the same here, but doing that i get an error in this line `return getProduct(state, bagItem?.product);` because bagItem can be undefined and the `getProduct` always expects a number.

Since it's not my area, I wasn't confortable changing this, so I used that as a workaround. But I agree with you, I don't like this approach too. I was thinking doing something like this:

`return getProduct(state, bagItem?.product || 0);`

Which fixes the error, but I'm afraid that it breaks something. What do you think?

mindouro: I also think it will break something ahah. Could you please add an "@TODO" on that selector so my team knows there is work to do there? 🙏 Thanks!

packages/redux/src/entities/selectors/entity.ts
```diff
@@ -42,36 +40,27 @@ export const getEntities = <
...
  *
- * @returns {object|undefined} - The specific entity of the given id, undefined
+ * @returns - The specific entity of the given id, undefined
  * if it isn't found.
  */
-export const getEntityById = <
-  SchemaName extends keyof NonNullable<StateEntities>,
-  EntityId extends keyof NonNullable<NonNullable<StateEntities>[SchemaName]>,
->(
+export const getEntityById = <T>(
```
mindouro, line 50: Why change this to "<T>"? You don't seem to be using it anywhere 🤔. This PR is a little confusing, you say you're "fixing" the selector and its errors but I can't find what those problems are 😕

packages/redux/src/bags/selectors.ts
andreecosta: Yes no problem :)

packages/redux/src/entities/selectors/entity.ts
andreecosta: The <T> means that it's using the type from the call. Which is being used in the return type. For example this call:

```javascript
export const getBrand = (
  state: StoreState,
  brandId: Brand['id'],
): Brand | undefined => getEntityById(state, 'brands', brandId);
```

It's passing the `Brand | undefined` as the response type for the getEntityById. Without this the response type was always unknown or any and every single call had an error because the type was unknown. This was the main problem with the function, the rest was just to clean it up a little bit.

nelsonleite (approved):

packages/redux/src/entities/selectors/entity.ts
mindouro: Got it, thanks!

PR #359, opened 2022-06-20
WIP: feat(redux): add patch checkout order item and delete checkout order item action creators

andrevechina:
## Description

Add action creators:
- patch checkout order item
- delete checkout order item 
And checkout order item entity mappers

## Checklist

- [ ] The commit message follows our guidelines
- [ ] Tests for the respective changes have been added
- [ ] The code is commented, particularly in hard-to-understand areas
- [ ] The labels and/or milestones were added

## Disclaimer

By sending us your contributions, you are agreeing that your contribution is made subject to the terms of our [Contributor Ownership Statement](https://github.com/Farfetch/.github/blob/master/COS.md)

PR #363, opened 2022-06-22
feat: add client and redux for `grouping`

mindouro:
## Description
This adds a client for the `products/{id}/grouping` endpoint.
This adds action, reducer, and selectors for the product `grouping`.

## Checklist

## Disclaimer


packages/client/src/products/types/productGrouping.types.ts
```diff
@@ -0,0 +1,31 @@
+export type DigitalAsset = {
+  mediaType: string;
+  displayOrder: number;
+  size: string;
+  url: string;
+  type: number;
```
anapim, line 6: Could this be an enum? According to the api documentation:
```
DigitalAssetType

0 - Default
1 - ItemSwatch
```

packages/client/src/products/types/productGrouping.types.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export type ProductGroupingEntry = {
+  digitalAssets: DigitalAsset[];
+  variationProperties: Array<{
+    type: 'Volume' | 'COLOR';
```
anapim, line 12: I know this is probably what is returned by the api, but do you think this will be fixed and should we leave a comment here?

packages/client/src/products/types/productGrouping.types.ts
```diff
@@ -0,0 +1,31 @@
...
+      value: string;
+    };
+  }>;
+  slug: string;
+  hasStock: boolean;
+  id: number;
+  order: number;
+  isDefault: boolean;
+  variantId: string;
+};
```
anapim, line 24: What do you think of ordering the properties alphabetically?

anapim (changes requested):

packages/client/src/products/types/productGrouping.types.ts
mindouro: Nice, I created an enum for it, thanks!

packages/client/src/products/types/productGrouping.types.ts
mindouro: Yup, this is how the API returns the types and I'm not confident it will change in the near future :/ so I added a comment saying that values come straight from the API.

packages/redux/src/products/actions/factories/fetchProductGroupingFactory.ts
```diff
@@ -0,0 +1,73 @@
...
+import productSchema from '../../../entities/schemas/product';
+import type { Dispatch } from 'redux';
+import type {
+  GetProductGrouping,
+  GroupingQuery,
+  Product,
+  ProductGrouping,
+} from '@farfetch/blackout-client/products/types';
+
+/**
```
andrevechina, line 17: I believe this doc block is not referencing the right function.
I had a similar issue on a previous [PR](https://github.com/Farfetch/blackout/pull/359/files#diff-6c5b873d0c4d39b5ed2522b401cf59c486bae4465670c38b2179df52e5774b78)

andrevechina (approved):

packages/redux/src/products/actions/factories/fetchProductGroupingFactory.ts
mindouro: You're 100% right, moved this doc block to be above the function it documents!

packages/client/src/products/types/groupingQuery.types.ts
```diff
@@ -0,0 +1,4 @@
+export type GroupingQuery = {
```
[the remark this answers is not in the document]
mindouro: That's a really cool tip but for this scenario I don't think it is worth adding comments 🙏

PR #406, opened 2022-07-15
fix(core|redux): fix hash creation for pagination props and commerce pages

vinicaval:
## Description

fix hash creation for pagination props and commerce pages:

- removed subfolder and queryString on slug for commercepages hash on serverinitialstate
- added page property with default value on hash builder
- changed hash format to '${contentType}!${codes}!${page}${pageSize}'
- pageSize property is optional and if its have value will be split by commas like:
posts!all!1,5

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/content/hooks/useContentPage.js
+++ b/packages/react/src/content/hooks/useContentPage.js
@@ -35,7 +35,7 @@ export default (slug, contentType, strategy = 'default') => {
     }
 
     return {
-      codes: slug,
+      codes: slug.split('?')[0],
       contentTypeCode: 'content_pages',
     };
   }, [slug]);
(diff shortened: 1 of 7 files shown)


joaoprcosta (approved):

packages/react/src/content/hooks/useContentPage.js
```diff
@@ -51,7 +51,7 @@ export default (slug, contentType, strategy = 'default') => {
   }, [action, slug, contentType, strategy]);
 
   useEffect(() => {
-    !contentPage && fetchContent();
+    !contentPage && !isLoading && fetchContent();
```
joaoprcosta, line 54: Adding `isLoading` to the dependency injection is a fake solution to avoid re-rendering. The problem is that `fetchContent` is not memoized and not stored inside `useEffect`, causing multiple re-renders. The list should contain: `contentPage`, slug, and `fetchContent`.

vinicaval: Thanks for find it. It's done

PR #412, opened 2022-07-25
feat(react): refactor useWishlist

andreribeirosilva:
## Description

This PR is part of the work being developed in order to achieve consistent interfaces among all the react hooks (provided in the `blackout-react` package).

The `useWishlist` hook is now exposing the following interface:
- Params
    - enableAutoFetch (optional)
- Return
    - isLoading
    - error
    - isFetched
    - data
        - id
        - count
        - items
        - isEmpty
    - actions
        - fetch()
        - reset()
        - addItem(productId, {sizeId, quantity})
        - updateItem(wishlistItemId, {sizeId, quantity})
        - removeItem(wishlistItemId) 

The `useWishlistItem` hook is now exposing the following interface:
- Params
    - enableAutoFetch (optional)
- Return
    - isLoading
    - error
    - isFetched
    - data
        - id
        - quantity
        - size
        - product
        - ...
    - actions
        - update({sizeId, quantity})
        - remove() 

### Dependencies

Depends on #411 

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/wishlists/hooks/__tests__/useWishlist.test.tsx
+++ b/packages/react/src/wishlists/hooks/__tests__/useWishlist.test.tsx
@@ -1,16 +1,46 @@
 import { cleanup, renderHook } from '@testing-library/react';
-import { mockStore } from '../../../../tests/helpers';
-import { mockWishlistState } from 'tests/__fixtures__/wishlists';
-import { Provider } from 'react-redux';
-import React from 'react';
+import {
+  mockWishlistId,
+  mockWishlistItemId,
+  mockWishlistState,
+} from 'tests/__fixtures__/wishlists';
+import { withStore } from '../../../../tests/helpers';
 import useWishlist from '../useWishlist';
+import type { StoreState } from '@farfetch/blackout-redux';
+
+const mockUserState = {
+  entities: {
+    user: {
+      id: 1234,
+      wishlistId: mockWishlistId,
+    },
+  },
+  users: {
+    error: null,
+    isLoading: false,
+    result: 1234,
+  },
+};
+
+const stateMockData: StoreState = {
+  ...mockWishlistState,
+  ...mockUserState,
+  entities: {
+    ...mockWishlistState.entities,
+    ...mockUserState.entities,
+  },
+};
 
 jest.mock('@farfetch/blackout-redux', () => ({
   ...jest.requireActual('@farfetch/blackout-redux'),
-  addWishlistItem: jest.fn(() => ({ type: 'addItem' })),
-  fetchWishlist: jest.fn(() => ({ type: 'fetch' })),
+  addWishlistItem: jest.fn(data => ({ type: 'add-item', ...data })),
+  updateWishlistItem: jest.fn((wishlistItemId, data) => ({
+    type: 'update-item',
+    wishlistItemId,
+    ...data,
+  })),
+  fetchWishlist: jest.fn(wishlistId => ({ type: 'fetch', wishlistId })),
   resetWishlist: jest.fn(() => ({ type: 'reset' })),
-  resetWishlistState: jest.fn(() => ({ type: 'resetState' })),
 }));
 
 const mockDispatch = jest.fn();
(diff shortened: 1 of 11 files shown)


packages/redux/src/entities/selectors/user.ts
```diff
@@ -102,6 +102,18 @@ export const getUserGender = (user: any): any => get(user, 'gender');
...
 
+/**
+ * Return the user's bag id.
+ *
+ * @param user - User data.
+ *
+ * @returns Bag id.
+ */
+export const getUserWishlistId = (user: any): any => {
+  console.log(user);
```
boliveira, line 113: Lingering console.log :)

packages/react/src/wishlists/hooks/__tests__/useWishlist.test.tsx
```diff
@@ -20,129 +49,227 @@ jest.mock('react-redux', () => ({
...
     });
 
-    expect(isLoading).toBe(true);
+    expect(isLoading).toBe(false);
   });
 
   describe('actions', () => {
-    it('should call `addWishlistItem` action', () => {
-      const { addWishlistItem } = getRenderedHook();
+    it('should call `fetch` action', () => {
```
boliveira, line 171: What do you think of creating a specific describe block for testing the options of this hook? Something like this:

```
describe('options', () => {
  describe('enableAutoFetch', () => {
    it('should call `fetch` if it is true, () => {});

    it('should not call `fetch` if it is false, () => {});
  });

});

```

packages/react/src/wishlists/hooks/__tests__/useWishlistItem.test.tsx
```diff
@@ -40,11 +40,14 @@ describe('useWishlistItem', () => {
...
-      updateWishlistItem: expect.any(Function),
       error: expect.any(Object),
       isLoading: expect.any(Boolean),
-      item: expect.any(Object),
+      isFetched: expect.any(Boolean),
+      actions: {
+        update: expect.any(Function),
+        remove: expect.any(Function),
+      },
+      data: expect.any(Object),
```
boliveira, line 50: Same as with the `useWishlist` hook, instead of expecting any object to be in data, what do you think of checking for the exact value from the mock state that you expect?

packages/react/src/wishlists/hooks/types/useWishlist.ts
```diff
@@ -1,29 +1,34 @@
...
-  isEmpty: boolean;
-  isLoading: boolean;
-  itemCount: number;
-  items: WishlistItemHydrated[] | undefined;
-  itemsWithSetsInfo: WishlistItemHydrated[] | undefined;
-  resetWishlist: () => void;
-  resetWishlistState: (fieldsToReset?: string[]) => void;
-  totalQuantity: number;
-  wishlist: WishlistsState['result'];
+export type Options = {
```
boliveira, line 9: This type is too general, what do you think of calling it `UseWishlistOptions`?

boliveira (changes requested):

packages/react/src/wishlists/hooks/useWishlist.ts
```diff
@@ -3,105 +3,78 @@
...
+  const fetch: FetchWishlist = useAction(fetchWishlist);
+  const reset: ResetWishlist = useAction(resetWishlist);
   // Data with some logic
-  const isEmpty = totalQuantity === 0;
-  // If there is no wishlist and there is no error (with `isWishlistLoading === false`),
-  // it means the wishlist hasn't started fetching, so it's considered loading
-  const isLoading = (!wishlist && !error) || isWishlistLoading;
+  const count = wishlist.count;
+  const isEmpty = count === 0;
+  const isFetched = !!wishlist.id && typeof count !== 'undefined' && count >= 0;
```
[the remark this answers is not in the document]
andreribeirosilva: This validation is necessary since TS throws an error identifying `count` as possibly `undefined`. If you prefere I can change it to `const isFetched = !!wishlist.id && !!count && count >= 0;`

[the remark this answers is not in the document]
andreribeirosilva: Actually, since after fetch we have both wishlist id and count, we don't need to validate the count value. 
Updated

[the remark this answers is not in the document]
boliveira: Nice! Thanks!

boliveira (approved):

PR #414, opened 2022-07-26
fix(react): fix GA4 integration not firing PRODUCT_UPDATED.CHANGE_QUANTITY when quantity is 0

AceBunny:
## Description

While implementing GA4 integration we received feedback that we should fire event `change_quantity` "when a user removes the last item from a product from the cart". Meaning firing `eventTypes.PRODUCT_UPDATED` with `oldQuantity` as previous added quantity and `quantity` as 0.

Since 0 is falsy, it will never fire this event because of the verification. So instead I think using lodash's `isFinite` is a better verification than just falsy check, as this makes it possible to fire `change_quantity` with `quantity` 0.

### Dependencies

None

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/analytics/integrations/GA4/commands.js
+++ b/packages/react/src/analytics/integrations/GA4/commands.js
@@ -12,6 +12,7 @@ import ga4EventNameMapping, {
   getEventProperties,
   InternalEventTypes,
 } from './eventMapping';
+import isFinite from 'lodash/isFinite';
 
 const genericCommandsBuilder = data => {
   const eventName = ga4EventNameMapping[data.event];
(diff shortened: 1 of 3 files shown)


ruifcnunes:
hello @AceBunny!
 first of all, thank you for your contribution :) 
 
 could you please add a unit test to cover this specific case, so we can proceed with the PR?
 
 thank you!

AceBunny:
Hi @ruifcnunes 

Just added a new united test for this case, can you check if it is ok? Thanks 😄

ruifcnunes (approved):

boliveira (approved):

danielbento92 (approved):

PR #423, opened 2022-07-28
feat(react): refactor useProductDetails hook

PedroMiguelSS:
## Description

This PR refactors the `useProductDetails` hook to its new interface:

- Params
    - productId
    - enabled (optional)
- Return
    - isLoading
    - error
    - isFetched
    - data
        - id
        - sizes
        - sizesWithRemainingQuantity
        - images
        - … (product entity data)
        - isOutOfStock
        - isOneSize
    - actions
        - refetch()
        - reset()

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
+++ b/packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
@@ -8,7 +8,7 @@ import {
   mockProductsState,
 } from 'tests/__fixtures__/products';
 import { renderHook } from '@testing-library/react';
-import { withStore } from '@farfetch/blackout-react/tests/helpers';
+import { withStore } from '../../../../tests/helpers';
 import useProductDetails from '../useProductDetails';
 
 jest.mock('@farfetch/blackout-redux', () => ({
@@ -18,6 +18,8 @@ jest.mock('@farfetch/blackout-redux', () => ({
 }));
 
 describe('useProductDetails', () => {
+  beforeEach(jest.clearAllMocks);
+
   it('should return data correctly with initial state', () => {
     const { result } = renderHook(() => useProductDetails(mockProductId), {
       wrapper: withStore(mockProductsState),
@@ -141,7 +143,7 @@ describe('useProductDetails', () => {
         },
       );
 
-      expect(fetchProductDetails).not.toHaveBeenCalledWith();
+      expect(fetchProductDetails).not.toHaveBeenCalled();
     });
   });
 
@@ -157,11 +159,9 @@ describe('useProductDetails', () => {
         wrapper: withStore(mockProductsState),
       });
 
-      const productsToReset = [10000, 20000];
+      reset();
 
-      reset(productsToReset);
-
-      expect(resetProductDetails).toHaveBeenCalledWith(productsToReset);
+      expect(resetProductDetails).toHaveBeenCalledWith([mockProductId]);
     });
 
     it('should call `refetch` action successfully', () => {
--- a/packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
+++ b/packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
@@ -157,9 +157,11 @@ describe('useProductDetails', () => {
         wrapper: withStore(mockProductsState),
       });
 
-      reset();
+      const productsToReset = [10000, 20000];
 
-      expect(resetProductDetails).toHaveBeenCalled();
+      reset(productsToReset);
+
+      expect(resetProductDetails).toHaveBeenCalledWith(productsToReset);
     });
 
     it('should call `refetch` action successfully', () => {
--- a/packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
+++ b/packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
@@ -1,75 +1,186 @@
-import { fetchProductDetails } from '@farfetch/blackout-redux';
-import { mockProductId, mockProductsState } from 'tests/__fixtures__/products';
-import { mockStore } from '../../../../tests/helpers';
-import { Provider } from 'react-redux';
+import {
+  fetchProductDetails,
+  resetProductDetails,
+} from '@farfetch/blackout-redux';
+import {
+  mockProduct,
+  mockProductId,
+  mockProductsState,
+} from 'tests/__fixtures__/products';
 import { renderHook } from '@testing-library/react';
-import React from 'react';
+import { withStore } from '@farfetch/blackout-react/tests/helpers';
 import useProductDetails from '../useProductDetails';
 
 jest.mock('@farfetch/blackout-redux', () => ({
   ...jest.requireActual('@farfetch/blackout-redux'),
-  fetchProductDetails: jest.fn(() => ({ type: 'get' })),
+  fetchProductDetails: jest.fn(() => () => Promise.resolve()),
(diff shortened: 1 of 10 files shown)


packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
```diff
@@ -1,75 +1,186 @@
...
+  fetchProductDetails,
+  resetProductDetails,
+} from '@farfetch/blackout-redux';
+import {
+  mockProduct,
+  mockProductId,
+  mockProductsState,
+} from 'tests/__fixtures__/products';
 import { renderHook } from '@testing-library/react';
+import { withStore } from '@farfetch/blackout-react/tests/helpers';
```
boliveira, line 11: Suggestion: This could be a relative import.

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
```diff
@@ -1,75 +1,186 @@
...
-    ),
+describe('useProductDetails', () => {
+  it('should return data correctly with initial state', () => {
+    const { result } = renderHook(() => useProductDetails(mockProductId), {
+      wrapper: withStore(mockProductsState),
+    });
+
+    expect(result.current).toStrictEqual({
+      error: null,
+      isFetched: true,
```
boliveira, line 28: I think this should be false. I understand that when `isFetched` is true, we have product data. Is that something that our `isProductFetched` selector is missing?

packages/react/src/products/hooks/useProductDetails.ts
```diff
@@ -1,138 +1,86 @@
...
-  const promotions = useSelector((state: StoreState) =>
-    getProductPromotions(state, id),
+  const isOneSize = useSelector((state: StoreState) =>
+    isProductOneSize(state, productId),
   );
-  const availableSizes = useSelector((state: StoreState) =>
-    getAllProductSizesRemainingQuantity(state, id),
+
+  const refetch = useCallback(
+    () => fetch(productId, fetchQuery, fetchForceDispatch, fetchConfig),
```
boliveira, line 51: What do you think of only making the refetch if `isLoading` is not true?

packages/react/src/products/hooks/useProductDetails.ts
```diff
@@ -1,138 +1,86 @@
...
-     */
-    product,
-    /**
-     * Product promotions.
-     */
-    promotions,
+    data: !!product
+      ? {
+          ...product,
+          isOneSize,
```
boliveira, line 76: What do you think of having the logic for `isOneSize` and `isOutOfStock` in the normalizer for the fetchProductDetailsAction so that the value will be available in the product entity, even when this hook is not used?

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
PedroMiguelSS: I think the name of this test might be misleading. In this test, the `isFetched` must be true because the store already has product data. 
Instead of `should return data correctly with initial state`, what do you think of `should return data correctly when data is fetched`?

packages/react/src/products/hooks/useProductDetails.ts
PedroMiguelSS: IMO, this action shouldn't do any of that kind of "magic". When one uses this hook to refetch data, one can (and should) only do that if `isLoading` is false.
What are your thoughts on this?

boliveira: I understand but at least it should in development give a warning to the console so the developer can identify any bugs in that situation. What do you think?

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
boliveira: My mistake, I commented the wrong line, I will close this and comment the correct line :)

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
```diff
@@ -1,75 +1,186 @@
...
+      },
+    };
+
+    const { result } = renderHook(() => useProductDetails(mockProductId), {
+      wrapper: withStore(errorMockProductsState),
+    });
+
+    expect(result.current).toStrictEqual({
+      error: 'Error - Not loaded.',
+      isFetched: true,
```
boliveira, line 66: I think this should be false. I understand that when isFetched is true, we have product data. Is that something that our isProductFetched selector is missing? (This is on the correct line now :) )

packages/react/src/products/hooks/useProductDetails.ts
PedroMiguelSS: IMO, putting these flags in the product entity is unnecessary. If someone doesn't want to use the hook and want to use the selectors instead, these two flags can be built with the information provided. 
I think that putting them in the store, is "replicating" information. What do you guys think?

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
```diff
@@ -1,75 +1,188 @@
...
+
+    it('should not fetch data if `enableAutoFetch` option is false', () => {
+      renderHook(
+        () => useProductDetails(mockProductId, { enableAutoFetch: false }),
+        {
+          wrapper: withStore(mockProductsState),
+        },
+      );
+
+      expect(fetchProductDetails).not.toHaveBeenCalledWith();
```
andreribeirosilva, line 144:
```suggestion
      expect(fetchProductDetails).not.toHaveBeenCalled();
```

andreribeirosilva (approved):

packages/react/src/products/hooks/useProductDetails.ts
boliveira: My idea was to remove those selectors altogether and have the normalizer extrapolate those values so the consumers would look at those flags directly on the product entity. I am still unsure if that normalization should be done at the client instead of doing it at the redux action. What do you think?

jhairabreu (approved):

packages/react/src/products/hooks/useProductDetails.ts
boliveira: As discussed offline, we will not implement this change as it might bring inconsistencies with other hooks and we are not sure if they are going to bring any benefits for users. We might reevaluate it in the future if anything changes.

packages/react/src/products/hooks/useProductDetails.ts
boliveira: As discussed offline, this idea might be a little tricky to implement because those properties are computed values so it would be best to not have them on the redux store although it is not very ergonomic to use the selectors to get those properties when we already have the product entity which contains many properties that do not need a selector to obtain. It would be much more interesting to have the Product entity be an instance with some get accessors which would calculate those derived properties but it would be tricky to have them in redux, so let's keep it the way it is now.

packages/react/src/products/hooks/__tests__/useProductDetails.test.tsx
boliveira: As discussed offline, we will change the way the `isFetched` prop works in all hooks to make it work like the `isFetched` value in react-query as much as possible. There was a bug in the details reducer that prevented this to work so a fix was introduced in another [PR](https://github.com/Farfetch/blackout/pull/431). Closing it...

PR #426, opened 2022-07-29
fix(redux): add reducer and selectors to personal ids

andreecosta:
## Description

- This adds the missing reducer and selectors for the sub-area Personal Ids

### Dependencies

None.

## Checklist

## Disclaimer


packages/redux/src/users/personalIds/selectors.ts
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Returns the loading status of the personal ids area.
+ *
+ * @param state - Application state.
+ *
+ * @returns Loader status.
+ */
+export const areUserPersonalIdsLoading = (
+  state: StoreState,
+): UserPersonalIdsState['isLoading'] =>
```
boliveira, line 21: Suggestion: If the selector is simple like this one, you can remove the return type annotation as it should be inferred by the body expression. Review this selector and other selectors where that rule applies

packages/redux/src/users/personalIds/types/action.types.ts
```diff
@@ -28,6 +28,28 @@ export type FetchPersonalIdsAction =
   | FetchPersonalIdsFailureAction
   | FetchPersonalIdsSuccessAction;
 
+//
+// Fetch personal id
+//
+export interface FetchPersonalIdFailureAction extends Action {
```
boliveira, line 34: Nitpicking: It would be better if it was renamed to `FetchUserPersonalIdFailureAction` to be consistent. This applies for all action types here, not just for the failure one.

boliveira (changes requested):

packages/redux/src/users/personalIds/selectors.ts
andreecosta: done :)

packages/redux/src/users/personalIds/types/action.types.ts
andreecosta: done :)

boliveira (approved):

PR #431, opened 2022-08-02
fix(redux|react): fix fetched selectors

boliveira:
## Description

- This fixes all fetched selectors to make sure they all work the same way:
  - If data has been fetched successfully, return true;
  - If fetch returned an error, return true;
  - If isLoading is true, return false;
This will bring it closer to the way that `react-query` works as they
defined the `isFetched` property similarly,
although once `isFetched` is true, it will not change to false if
another fetch is made after, which is not the case here. To make it work
the same way as `react-query` would be unfeasible right now,
so this is an acceptable trade-off.

- Added `isBagFetched` and `isWishlistFetched` selectors and use them
in the `useBag` and `useWishlist` hooks respectively for the calculation
of its `isFetched` properties.

## Checklist

## Disclaimer


mindouro (approved):

packages/redux/src/wishlists/selectors/__tests__/wishlists.test.ts
```diff
@@ -64,6 +64,34 @@ describe('wishlists redux selectors', () => {
     });
   });
 
+  describe('isWishlistFetched()', () => {
+    it('should return true if the bag fetch request succeeded', () => {
```
jhairabreu, line 68: shouldn't this be 

...if the `wishlist` fetch request... ?

packages/redux/src/wishlists/selectors/__tests__/wishlists.test.ts
```diff
@@ -64,6 +64,34 @@ describe('wishlists redux selectors', () => {
...
+      expect(selectors.isWishlistFetched(mockWishlistState)).toBe(true);
+    });
+    it('should return true if the bag fetch request failed', () => {
```
jhairabreu, line 71: Same here

jhairabreu (changes requested):

jhairabreu (approved):

PR #447, opened 2022-08-05
chore(redux): move entity selectors from entities folder

boliveira:
## Description

- This moves all entities selectors that are specific to an area
to its area's corresponding folder. This is to avoid leaking
implementation details of how the redux package works to the
consumer and thus having a much more consistent and natural import
pattern. For example, to import the `getProduct` selector,
you just need to import it from the `products` folder as it now
lives alongside the other product selectors. This, of course,
only applies for consumers that import our packages directly from
a folder instead of importing from the root.

BREAKING CHANGE: All entity selectors must now be imported from its
respective area instead of being import from the `entities` folder.

## Checklist

## Disclaimer


packages/redux/src/brands/selectors.ts
```diff
@@ -63,6 +63,30 @@ export const areBrandsLoading = (
...
+ *
+ * @param state   - Application state.
+ * @param brandId - Brand id.
+ *
+ * @returns Brand normalized.
+ */
+export const getBrand = (
+  state: StoreState,
+  brandId: Brand['id'],
+): Brand | undefined => getEntityById(state, 'brands', brandId);
```
jhairabreu, line 88: Do we need this `: Brand | undefined` return type?

packages/redux/src/categories/selectors/categories.ts
```diff
@@ -58,3 +60,37 @@ export const areCategoriesLoading = (state: StoreState) =>
...
+ *
+ * @param state      - Application state.
+ * @param categoryId - Category id.
+ *
+ * @returns Category normalized or undefined is nothing found.
+ */
+export const getCategory = (
+  state: StoreState,
+  categoryId: Category['id'],
+): Category | undefined => getEntityById(state, 'categories', categoryId);
```
jhairabreu, line 75: Same here. Check for the rest of your changes also

jhairabreu (changes requested):

packages/redux/src/brands/selectors.ts
boliveira: No, we don't. I was not looking at this when making these changes but I reviewed all the selectors I have worked with on this PR and removed any unnecessary typings. Can you take a look at it? Thanks!!

packages/redux/src/categories/selectors/categories.ts
boliveira: Reviewed typing for all the selectors on this PR. Can you take a look? Thanks!

packages/redux/src/categories/selectors/categories.ts
```diff
@@ -58,3 +60,33 @@ export const areCategoriesLoading = (state: StoreState) =>
...
 export const areCategoriesFetched = (state: StoreState) =>
   getIsFetched(state.categories as CategoriesState);
+
+/**
+ * Returns a specific category by its id.
+ *
+ * @param state      - Application state.
+ * @param categoryId - Category id.
+ *
+ * @returns Category normalized or undefined is nothing found.
```
jhairabreu, line 70: Should be `if` instead of `is` in `undefined is nothing found.`

jhairabreu (changes requested):

giuliamml (approved):

jhairabreu (approved):

PR #450, opened 2022-08-08
feat(react): refactor useProductListing hook

PedroMiguelSS:
## Description

This PR refactors the `useProductListing` hook to its new interface:

- Params
    - slug
    - options: (query, setProductsListingHash, type, useCache, enabled)
- Return
    - isLoading
    - error
    - isFetched
    - data
        - id
        - breadcrumbs
        - … (product entity data)
        - pagination
            - pageIndex
            - totalItems
            - totalPages
        - items
    - actions
        - refetch()
        - reset()

### Dependencies

None.

## Checklist

## Disclaimer


packages/react/src/products/hooks/types/useProductListing.ts
```diff
@@ -0,0 +1,37 @@
+import type {
+  Config,
+  GetProductListingQuery,
+  GetProductSetQuery,
+} from '@farfetch/blackout-client';
+
+export enum ProductListingTypes {
+  Listing = 'listing',
```
jhairabreu, line 8: For consistency, can you change this `listing` to `Listing`?

jhairabreu (changes requested):

packages/react/src/products/hooks/useProductListing.ts
```diff
@@ -0,0 +1,116 @@
...
+      setProductsListHash,
+      slug,
+      useCache,
+    ],
+  );
+
+  const shouldLoadListing = enableAutoFetch && !isLoading && !error && !listing;
+
+  useEffect(() => {
+    shouldLoadListing && fetch();
```
andreribeirosilva, line 95: I believe that calling the fetch conditionally (`enableAutoFetch && !isLoading && !error && !listing;`) we'll not be able to `setProductsListHash` if data is cached. Should we remove the `!listing` from the `shouldLoadListing` and let the action `fetchListing` deal with the cache mechanism? 

Or am I missing something here? 🤔

andreribeirosilva (changes requested):

packages/react/src/products/hooks/useProductListing.ts
PedroMiguelSS: Yes, I believe we should do that change. We have discussed this previously but I forgot to do this fix 👍

jhairabreu (approved):

packages/react/src/products/hooks/useProductListing.ts
PedroMiguelSS: As we have discussed, I'll let the `!listing` in the condition above and we will remove both `setListingHash` and `useCache` logic later in [this ticket](https://trello.com/c/qPHzju7c/29-refactor-productlisting-reducer-to-remove-hash).

boliveira (approved):

PR #467, opened 2022-08-26
fix(react): add products brand data to useProductListing hook

PedroMiguelSS:
## Description

This PR:
- adds brand data to each product returned by the `useProductListing` hook;
- fixes listing `serverInitialState` to infer whether the page is of type `set` or not. This change fixes the fact that the listing hash was being miscalculated (it was always calculated as `listing/<something>` instead of `sets/<something>` when it was a set page).

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/products/hooks/useProductListing.ts
+++ b/packages/react/src/products/hooks/useProductListing.ts
@@ -103,6 +103,7 @@ const useProductListing = (
       ? {
           ...listing,
           pagination,
+          hash: productListingHash,
           items: products,
         }
       : undefined,
--- a/packages/redux/src/products/serverInitialState/lists.ts
+++ b/packages/redux/src/products/serverInitialState/lists.ts
@@ -46,7 +46,7 @@ const serverInitialState: ListsServerInitialState = ({
   delete query.json;
 
   const builtSlug = getSlug(pathname);
-  const isSet = model?.pageType === 'sets';
+  const isSet = model?.pageType === 'set';
   const hash = generateProductsListHash(builtSlug, query, { isSet });
 
   // Normalize it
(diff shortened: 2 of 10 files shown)


andreribeirosilva (approved):
LGTM

packages/react/src/products/hooks/useProductListing.ts
```diff
@@ -62,6 +63,17 @@ const useProductListing = (
     getProductsListProducts(state, productListingHash),
   );
 
+  const productsWithBrands = useSelector((state: StoreState) =>
```
nelsonleite, line 66: What if this was a redux selector as well, would it be useful?

anapim:
Please update milestone.

packages/redux/src/products/serverInitialState/lists.ts
```diff
@@ -46,7 +46,7 @@ const serverInitialState: ListsServerInitialState = ({
   delete query.json;
 
   const builtSlug = getSlug(pathname);
-  const isSet = model?.pageType === 'sets';
+  const isSet = model?.pageType === 'set';
```
nelsonleite, line 49: Question/suggestion - instead of a hardcoded "set" value, what about getting it from the TS enum of the model type? I mean something like `Model['pageType'].Set` (not sure if it's correct though)

catarinacouto (approved):

packages/react/src/products/hooks/useProductListing.ts
PedroMiguelSS: This was already a selector: `getBrand`, that was being used inside the hook.   
But I think I did understand what you meant. I've updated the `getProductsListProducts` selector to return the product's data instead of returning just the brand id.
What do you think?

packages/redux/src/products/serverInitialState/lists.ts
nelsonleite: I don't really know 🤔 

I'll try to understand it and get back to you, in the meantime I'm resolving this, no need to block because of it.

nelsonleite (approved):

PR #470, opened 2022-08-29
feat(core): add new selector for order splitting

pedro-gbf:
## Description

This adds a new selector for order splitting in the stable edition. This will provide a more optimised organization of the data in the store to be rendered in the tenant.

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/orders/redux/selectors.js
+++ b/packages/core/src/orders/redux/selectors.js
@@ -573,3 +573,47 @@ export const isOrderItemAvailableActivitiesLoading = state =>
  */
 export const getOrderItemAvailableActivitiesError = state =>
   getOrderItemAvailableActivities(state.orders).error;
+
+/**
+ * Returns a list of order summaries, with the order items if the
+ * details of the order were previously fetched.
+ *
+ * This selector should only be used when implementing splitted orders.
+ * That depends on the prop `splitByMerchantOrderCode` to be true
+ * when fetching the user orders in the `doGetOrders` action.
+ *
+ * @function
+ * @param {string} orderId - Order id.
+ * @param {object} state - Application state.
+ * @returns {object} Order summaries array.
+ */
+export const getOrderShipments = (state, orderId) => {
+  const order = getOrder(state, orderId);
+
+  if (!order) return;
+
+  // Omit both byMerchant and items, we don't want to expose all of the items
+  // of the order by its alphanumeric, we split the order items by the order summary
+  const { byMerchant, items, ...sharedOrderInformation } = order;
+  const result = [];
+
+  for (const merchantId in byMerchant) {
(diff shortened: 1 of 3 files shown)


jpklzm (approved):

packages/core/src/orders/redux/selectors.js
```diff
@@ -573,3 +573,42 @@ export const isOrderItemAvailableActivitiesLoading = state =>
  */
 export const getOrderItemAvailableActivitiesError = state =>
   getOrderItemAvailableActivities(state.orders).error;
+
+/**
+ * Returns the error for the order item available activities operations.
```
andreecosta, line 578: I think you forgot to change the description 🙂

andreecosta (changes requested):

packages/core/src/orders/redux/selectors.js
pedro-gbf: Indeed, changed it 👍

andreecosta (approved):

packages/core/src/orders/redux/selectors.js
```diff
@@ -573,3 +573,43 @@ export const isOrderItemAvailableActivitiesLoading = state =>
...
+
+  if (!order) return;
+
+  // Omit both byMerchant and items, we don't want to expose all of the items
+  // of the order by its alphanumeric, we split the order items by the order summary
+  const { byMerchant, items, ...sharedOrderInformation } = order;
+  const result = [];
+
+  for (const merchantId in byMerchant) {
+    for (const merchantOrderCode in byMerchant[merchantId]) {
```
boliveira, line 597: I think this name is misleading, as this variable will contain the keys inside the `byMerchant[merchantId]` object which is different from the `merchantOrderCode` property which exists inside the order object. What do you think of calling this just `key` or another more general name?

ruifcnunes (approved):

boliveira (changes requested):

packages/core/src/orders/redux/selectors.js
pedro-gbf: The name is correct, this selector only works for the order split flow. So when the `splitByMerchantOrderCode` inside of the `byMerchant[merchantId]` object there will be other objects using the merchantOrderCode as a key.

I’ll add an extra comment to warn the devs that this should only be used when the `doGetOrders` is provided the prop `splitByMerchantOrderCode` as true. You can find more about this inside of that action.

pedro-gbf: Added an extra comment in the selector comment, please check if it’s clearer.

boliveira: I see, I was working on the `next` branch which still does not have that split logic. Thanks!

boliveira (approved):

PR #474, opened 2022-09-01
feat(client|react|redux): add useProductGrouping and useProductGroupingProperties hooks

franciscofarinha:
## Description
- Add `useProductGrouping`
- Add `useProductGroupingProperties`
- Add `grouping` selectors
- Add `groupingProperties` selectors
- Add `getProductGroupingProperties` request
- Add `fetchProductGroupingPropertiesFactory` 
- Refactor `fetchProductGroupingFactory` to cache results
- Refactor `grouping` adapter to receive cached structure
- Add `groupingProperties` adapter
- Fix `buildQueryStringFromObject` to allow false values 

## Checklist

## Disclaimer


catarinacouto (approved):

packages/redux/src/helpers/buildQueryStringFromObject.ts
```diff
@@ -31,7 +31,8 @@ const buildQueryStringFromObject = (
     // which needs the `|` separating each different values of the same facet
     const parsedValue = Array.isArray(value) ? value.join('|') : value;
 
-    parsedValue && paramsToUrl.push([`${key.toLowerCase()}=${parsedValue}`]);
+    (parsedValue || typeof parsedValue === 'boolean') &&
```
nelsonleite, line 34:
```suggestion
    parsedValue !== undefined &&
```

Seems simpler, and I believe it accomplishes the same goal 🙂

packages/redux/src/products/reducer/groupingProperties.ts
```diff
@@ -0,0 +1,76 @@
...
+  switch (action.type) {
+    case actionTypes.FETCH_PRODUCT_GROUPING_PROPERTIES_REQUEST:
+      return {
+        ...state,
+        [action.meta.productId]: true,
+      };
+    case actionTypes.FETCH_PRODUCT_GROUPING_PROPERTIES_SUCCESS:
+      return { ...state, [action.meta.productId]: false };
+    case actionTypes.FETCH_PRODUCT_GROUPING_PROPERTIES_FAILURE:
+      return { ...state, [action.meta.productId]: undefined };
```
nelsonleite, line 50: Is the loading state supposed to be undefined on failure? This will make the `isProductGroupingLoading` selector return not only a boolean value, even after a request has been made 🤔 

(Seeing another reducer in a recent PR, [the `isLoading` is false on failure](https://github.com/Farfetch/blackout/blob/50455210f56c06007cd6a303e397660aa634078a/packages/redux/src/search/reducer/searchSuggestions.ts#L24))

If it's supposed, can you please explain the reasoning to have different approaches for me (and anyone who sees it) to understand? 🙏

packages/redux/src/products/selectors/grouping.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param state - Application state.
+ * @param id    - Product id.
+ *
+ * @returns If the product grouping is loading or not.
+ */
+export const isProductGroupingLoading = (
+  state: StoreState,
+  id: ProductEntity['id'],
+) => getIsLoading((state.products as ProductsState).grouping)[id];
```
nelsonleite, line 18: Why does this need to be cast `as ProductsState` in every occurrence? 🤔

andreribeirosilva (approved):

packages/react/src/products/hooks/useProductGrouping.ts
```diff
@@ -0,0 +1,65 @@
...
+) => {
+  const {
+    fetchConfig,
+    enableAutoFetch = true,
+    fetchQuery = { pageIndex: 1 },
+  } = options;
+  const fetchAction = useAction(fetchProductGrouping);
+
+  const error = useSelector((state: StoreState) =>
+    getProductGroupingError(state, productId),
```
boliveira, line 28: Shouldn't these selectors receive the query hash as well? If the fetchQuery changes between renders (or different components that are rendered at the same time use different query parameters), it will not reflect the state of the correct request. This is valid for `isProduuctGroupingLoading` and `isProductGroupingFetched` selectors as well.

packages/react/src/products/hooks/useProductGrouping.ts
```diff
@@ -0,0 +1,65 @@
...
+    isProductGroupingFetched(state, productId),
+  );
+  const queryString = fetchQuery && buildQueryStringFromObject(fetchQuery);
+  const hash = queryString ? queryString : '!all';
+  const productGrouping = useSelector((state: StoreState) =>
+    getProductGrouping(state, productId, hash),
+  );
+
+  const shouldLoadProductGrouping =
+    enableAutoFetch && !isLoading && !error && !productGrouping;
```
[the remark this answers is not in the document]
boliveira: Actually you only need this:

```suggestion
    enableAutoFetch && !isLoading && !isFetched;
```
This is because `isFetched` will be true when there is an error or a result in the store, but it will be false when the it is loading so we need to explicitly check that it is not loading as well.

packages/react/src/products/hooks/useProductGroupingProperties.ts
```diff
@@ -0,0 +1,61 @@
...
+import { useAction } from '../../helpers';
+import { useCallback, useEffect } from 'react';
+import { useSelector } from 'react-redux';
+import type { UseProductGroupingPropertiesOptions } from './types';
+
+const useProductGroupingProperties = (
+  productId: ProductEntity['id'],
+  options: UseProductGroupingPropertiesOptions = {},
+) => {
+  const { fetchConfig, enableAutoFetch = true, fetchQuery = {} } = options;
```
boliveira, line 20: Does `fetchQuery` needs to have a default value here? I think as the endpoint does not require it, you can remove the default value and avoid having the `fetchQuery` having a different object reference for each render, which might trigger the useEffect for fetching as it will create a new `fetch` reference each time the hook is rendered.

packages/react/src/products/hooks/useProductGroupingProperties.ts
```diff
@@ -0,0 +1,61 @@
...
+  const fetchAction = useAction(fetchProductGroupingProperties);
+
+  const error = useSelector((state: StoreState) =>
+    getProductGroupingPropertiesError(state, productId),
```
boliveira, line 24: Same as https://github.com/Farfetch/blackout/pull/474/files#r974009030

packages/react/src/products/hooks/__tests__/useProductGrouping.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+    });
+
+    it('should not fetch data if `enableAutoFetch` option is false', () => {
+      renderHook(
+        () =>
+          useProductGrouping(mockProductId, {
+            enableAutoFetch: false,
+          }),
+        {
+          wrapper: withStore(mockProductsState),
```
boliveira, line 128: This state corresponds to an already loaded state, which means that even if `enableAutoFetch` was true, it would not make a fetch request. To be correct, this test should use a state where the product grouping is not loaded yet to make sure that the fetch request was not made because `enableAutoFetch` was false and not that data was already loaded.

packages/react/src/products/hooks/__tests__/useProductGroupingProperties.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+    it('should not fetch data if `enableAutoFetch` option is false', () => {
+      renderHook(
+        () =>
+          useProductGroupingProperties(mockProductId, {
+            enableAutoFetch: false,
+          }),
+        {
+          wrapper: withStore(mockProductsState),
```
boliveira, line 137: Same as https://github.com/Farfetch/blackout/pull/474/files#r974041196

packages/redux/src/products/actions/factories/fetchProductGroupingFactory.ts
```diff
@@ -37,19 +40,29 @@ const fetchProductGroupingFactory =
...
+      const product = getProduct(state, productId);
+      const previousGroupings = product?.grouping;
+      const queryString = query && buildQueryStringFromObject(query);
+      const hash = queryString ? queryString : '!all';
 
-      // Replace all the grouping data with the new one
       const productWithGrouping = {
         id: productId,
-        grouping: result,
+        grouping: {
```
boliveira, line 62: Instead of doing this, what do you think of using an entity mapper for merging the results in the store? If you use an entity mapper for the `FETCH_PRODUCT_GROUPING_SUCCESS` action, you can perform the logic to merge all the groupings instead of having the action factory retrieve the previous groupings from the state and merge them.

packages/redux/src/products/reducer/groupingProperties.ts
franciscofarinha: You're right, it's not supposed to be undefined on failure. I based it on the grouping reducer. I'll fix it here and there as well

packages/react/src/products/hooks/useProductGrouping.ts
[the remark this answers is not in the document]
franciscofarinha: I'll change it

packages/redux/src/helpers/buildQueryStringFromObject.ts
franciscofarinha: I can't check if it's not undefined because it will allow empty strings and other falsy values that are defined

nelsonleite (approved):

packages/redux/src/products/selectors/grouping.ts
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @returns If the product grouping is loading or not.
+ */
+export const isProductGroupingLoading = (
+  state: StoreState,
+  id: ProductEntity['id'],
+  hash: string,
+) => {
+  const grouping = state.products?.grouping;
+  return grouping ? getIsLoading(grouping)[id]?.[hash] : undefined;
```
boliveira, line 19: You can avoid the check for the grouping property. Assume that the grouping property will always be available on the redux state. If the developer calls this method and did not provide the `grouping` reducer inside the `product` reducer, it will throw an error so the caller can properly know that it needs to configure the `grouping` reducer instead of returning `undefined`.

You can do it like this:

```suggestion
   return getIsLoading((state.products as ProductsState).grouping)[id]?.[hash];
```

PR #497, opened 2022-09-16
feat(core): add omnitracking's product updated events mappings

danielbento92:
## Description

- Added product updated (color, size, quantity) event mapping to omnitracking;
- Added feature requirements support to bag and wishlist middlewares.

### Dependencies

None.

## Checklist

- [x] The commit message follows our guidelines
- [x] Tests for the respective changes have been added
- [x] The code is commented, particularly in hard-to-understand areas
- [x] The labels and/or milestones were added

## Disclaimer

By sending us your contributions, you are agreeing that your contribution is made subject to the terms of our [Contributor Ownership Statement](https://github.com/Farfetch/.github/blob/master/COS.md)

PR #498, opened 2022-09-19
Feat add omnitracking product clicked event

danielbento92:
## Description

Added product clicked event mapping to omnitracking;

### Dependencies

#497.

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/analytics/integrations/Omnitracking/definitions.js
+++ b/packages/core/src/analytics/integrations/Omnitracking/definitions.js
@@ -8,6 +8,7 @@ import {
   getCheckoutEventGenericProperties,
   getCommonCheckoutStepTrackingData,
   getGenderValueFromProperties,
+  getOmnitrackingProductId,
   getProductLineItems,
   getProductLineItemsQuantity,
   getValParameterForEvent,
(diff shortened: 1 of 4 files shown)


packages/core/src/analytics/integrations/Omnitracking/definitions.js
```diff
@@ -523,6 +524,88 @@ export const trackEventsMapper = {
     wishlistId: data.properties?.wishlistId,
     lineItems: getProductLineItems(data),
   }),
+  [eventTypes.PRODUCT_CLICKED]: data => ({
+    tid: 2926,
```
ruifcnunes, line 528: I think we could also map the productId here, like in the PRODUCT_UPDATED mapping. what do you think?

danielbento92: ProductId field is inside of lineItems property. 
LineItems it's a json string, with product clicked information.

boliveira (approved):

packages/core/src/analytics/integrations/Omnitracking/definitions.js
danielbento92: Added product id to event track property list.

ruifcnunes (approved):

PR #500, opened 2022-09-19
fix(react): remove `useAction` from public exports

### LINKED ISSUE #401: useAction Hook is obligating to install redux-related packages when using blackout-react
## Expected behaviour

When using the helpers from `blackout-react`, no redux-related packages should be required to be installed, if we use no features this package which don't require redux explicity:

```js
import { headers, getTimeInMinutes } from '@farfetch/blackout-react`;

...
```

## Actual behaviour

When we import helpers, like this:
```js
import { headers, getTimeInMinutes } from '@farfetch/blackout-react`;

...
```
as the `@farfetch/blackout-react` exports `useAction` on the `helpers/index.ts`, it requires the installation of `redux` and `react-redux`.

## Steps to reproduce

1. Open a project without redux and react-redux installed
2. Try to make the following import
```js
import { headers, getTimeInMinutes } from '@farfetch/blackout-react`;
```
3. Run the project

## Context/environment

This happens on projects that have installed `@farfetch/blackout-react` but doesn't have `redux` and `react-redux` and need to use some react helpers.

## Possible fix

1. Don't export `helpers/useAction.ts` directly on the `helpers/index.ts`, and require users to use a explicitly path, like: 
```js
import { useAction } from @farfetch/blackout-react/helpers/withRedux
// or
import useAction from @farfetch/blackout-react/helpers/useAction
```

2. Remove `useAction`, as this function is known for having performance issues and is not recommended (see https://react-redux.js.org/api/hooks#recipe-useactions) and pass the decision to use that for the tenant

boliveira:
## Description

- This removes the `useAction` module from the public exports
of the react package as it is an internal helper module and
was requiring the installation of the redux package even when
not importing this module specifically (ex: importing another
module from the package root).

BREAKING CHANGE: `useAction` hook is not available from
react package. Use `useDispatch` hook from `react-redux` directly instead.

Closes #401 

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/__tests__/__snapshots__/index.test.ts.snap
+++ b/packages/react/src/__tests__/__snapshots__/index.test.ts.snap
@@ -172,7 +172,6 @@ Object {
     "Mobile": "mobile",
     "Web": "web",
   },
-  "useAction": [Function],
   "useBag": [Function],
   "useBagItem": [Function],
   "useCategories": [Function],
--- a/packages/react/src/bags/hooks/useBag.ts
+++ b/packages/react/src/bags/hooks/useBag.ts
@@ -21,9 +21,9 @@ import {
   updateBagItem as updateBagItemAction,
 } from '@farfetch/blackout-redux';
 import { ProductError, SizeError } from './errors';
-import { useAction } from '../../helpers';
 import { useCallback, useEffect } from 'react';
 import { useSelector, useStore } from 'react-redux';
+import useAction from '../../helpers/useAction';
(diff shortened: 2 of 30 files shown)


jhairabreu (changes requested):
Can you leave a link to the issue for this somewhere in the commit message?

other than that, looks good to me

jhairabreu (approved):

giuliamml (approved):

catarinacouto (approved):

PR #502, opened 2022-09-19
feat(core|react): share uuid between omnitracking and google analytics 4

ruifcnunes:
## Description
This PR adds a new feature to enable the identification of an event between integrations. The calculation is done on the analytics core layer and shared between integrations via the event context. For now only the omnitracking and google analytics map this new property, but it's ready to be mapped in other integrations if necessary.

### Dependencies
None.

## Checklist

## Disclaimer


packages/core/src/analytics/__tests__/analytics.test.js
```diff
@@ -776,7 +776,13 @@ describe('analytics', () => {
             type: trackTypes.TRACK,
             event,
             properties,
-            context,
+            context: {
+              ...context,
+              event: {
+                ...context.event,
+                uuid: expect.any(String),
```
boliveira, line 783: Is the `expect.any` here correct?

ruifcnunes: yes, since this const is only used on the expected result, we must ensure the integration track spy is called with the `uuid` that analytics will create.

packages/react/src/analytics/integrations/GA4/commands.js
```diff
@@ -12,12 +12,18 @@ import ga4EventNameMapping, {
...
   InternalEventTypes,
 } from './eventMapping';
+import get from 'lodash/get';
 import isFinite from 'lodash/isFinite';
 
 const genericCommandsBuilder = data => {
   const eventName = ga4EventNameMapping[data.event];
+  const eventProperties = getEventProperties(data.event, data);
+  const eventData = {
+    uuid: get(data, 'context.event.uuid'),
```
boliveira, line 22: I think the `uuid` should be after the spread of the `eventProperties` here to avoid the case where the event properties contain a `uuid` parameter as well. Question: Should we have a different name for the `uuid` parameter since it is an injected parameter of the event? Something like `__uuid` to make it explicit that it is being injected by analytics. What do you think?

packages/core/src/analytics/__tests__/analytics.test.js
boliveira: Oh you're right, I thought `data` was being used somewhere else as well. Resolving it. Thanks!!

danielbento92 (approved):

PR #503, opened 2022-09-20
Feat add omnitracking interact content event

danielbento92:
## Description

- Added interact content event mapping to omnitracking;

### Dependencies

#501.

## Checklist

- [x] The commit message follows our guidelines
- [x] Tests for the respective changes have been added
- [x] The code is commented, particularly in hard-to-understand areas
- [x] The labels and/or milestones were added

## Disclaimer

By sending us your contributions, you are agreeing that your contribution is made subject to the terms of our [Contributor Ownership Statement](https://github.com/Farfetch/.github/blob/master/COS.md)

PR #504, opened 2022-09-20
Feat add omnitracking select content event

danielbento92:
## Description

- Added select content event mapping to omnitracking;
- Reafactoring of interact content tests.

### Dependencies

#503

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/analytics/integrations/__tests__/Omnitracking.test.js
+++ b/packages/core/src/analytics/integrations/__tests__/Omnitracking.test.js
@@ -342,6 +342,9 @@ describe('Omnitracking', () => {
 
           expect(mockLoggerWarn).not.toHaveBeenCalled();
         });
+      });
+
+      describe('Event Tracking', () => {
         it('should warn if an interact content does not have required parameters', async () => {
           const data = generateTrackMockData({
             event: eventTypes.INTERACT_CONTENT,
@@ -354,6 +357,21 @@ describe('Omnitracking', () => {
           expect(mockLoggerWarn).toHaveBeenCalledWith(
             expect.stringContaining(
               'properties "contentType" and "interactionType" should be sent',
+            ),
+          );
+        });
+        it('should warn if an select content does not have required parameters', async () => {
+          const data = generateTrackMockData({
+            event: eventTypes.SELECT_CONTENT,
+            properties: {
+              contentType: undefined,
+            },
+          });
+          await omnitracking.track(data);
+
+          expect(mockLoggerWarn).toHaveBeenCalledWith(
+            expect.stringContaining(
+              'properties "contentType" and "id" should be sent',
             ),
           );
         });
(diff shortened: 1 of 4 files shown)


ruifcnunes (approved):

packages/core/src/analytics/integrations/__tests__/Omnitracking.test.js
```diff
@@ -372,7 +373,96 @@ describe('Omnitracking', () => {
...
+          it('should warn if an interact content does not have required parameters', async () => {
+            const data = generateTrackMockData({
+              event: eventTypes.INTERACT_CONTENT,
+              properties: {
+                interactionType: undefined,
+              },
+            });
+            await omnitracking.track(data);
+
+            expect(mockLoggerWarn).toHaveBeenCalledWith(
```
boliveira, line 388: I would also add an `expect` to make sure that the event is not being sent to omnitracking in this case. If you do that, change the test description to something like this: "should not track an interact content event if the required parameters are missing" or something like that.

packages/core/src/analytics/integrations/__tests__/Omnitracking.test.js
```diff
@@ -372,7 +373,96 @@ describe('Omnitracking', () => {
...
+              expect.objectContaining({
+                parameters: expect.objectContaining({
+                  tid: 668,
+                }),
+              }),
+            );
+          });
+        });
+
+        it('should not track any event', async () => {
```
boliveira, line 416: This test description is too generic, I think it would read best if it was something like "should not track an event when `interactionType` is scroll but target is not `document.body`". What do you think?

packages/core/src/analytics/integrations/__tests__/Omnitracking.test.js
```diff
@@ -372,7 +373,96 @@ describe('Omnitracking', () => {
...
+              contentType: 'logo',
+              id: 'home_logo',
+            },
+          });
+          await omnitracking.track(data);
+
+          expect(postTrackingsSpy).toHaveBeenCalledWith(
+            expect.objectContaining({
+              parameters: expect.objectContaining({
+                tid: 2882,
```
boliveira, line 444: Missing the remaining parameters: `contentType`, `interactionType`, `val` and `actionArea`. Use an event that will generate an omnitracking payload that contain all of those properties, to make sure the test cover it.

boliveira (approved):

packages/core/src/analytics/integrations/__tests__/Omnitracking.test.js
danielbento92: This parameters will be tested with iteration of each event (using fixtures data), that should be equal than a snapshot with these fields.

boliveira: Yes, I was looking at the latest commit only which did not have changes on the snapshots but I see that it was in a previous commit relative to a previous PR.

PR #509, opened 2022-09-21
feat(redux|client|react): add categories hooks

JoaoFerreira10:
## Description

In this PR the categories reducers are refactored in order to allow to maintain a store cache.

This PR also adds to blackout-react the following categories hooks:

## useCategories

- Params
  - enableAutoFetch (optional)
- Return
  - isLoading
  - error
  - isFetched
  - data
  - actions
    -  fetch
    - reset 

## useCategory

- Params
  - id   
  - enableAutoFetch (optional)
- Return
  - isLoading
  - error
  - isFetched
  - data
  - actions
    -  fetch
    - refetch

## useTopCategories

- Params
  - enableAutoFetch (optional)
- Return
  - isLoading
  - error
  - isFetched
  - data
  - actions
    - fetch
    - reset 

## Checklist

## Disclaimer


packages/react/src/categories/utils.ts
```diff
@@ -0,0 +1,8 @@
+import type { Category } from '@farfetch/blackout-client';
+
+export const getRootCategory = (id: number, categories: Category[]) => {
```
catarinacouto, line 3: I believe we are missing unit tests for this util :)

packages/redux/src/categories/actionTypes.ts
```diff
@@ -29,7 +29,21 @@ export const FETCH_CATEGORIES_REQUEST =
  */
 export const FETCH_CATEGORIES_SUCCESS =
   '@farfetch/blackout-redux/FETCH_CATEGORIES_SUCCESS';
-
+/**
+ * Action type dispatched when the fetch categories request fails.
```
catarinacouto, line 33:
```suggestion
 * Action type dispatched when the fetch category request fails.
```

packages/redux/src/categories/actionTypes.ts
```diff
@@ -29,7 +29,21 @@ export const FETCH_CATEGORIES_REQUEST =
...
+ */
+export const FETCH_CATEGORY_FAILURE =
+  '@farfetch/blackout-redux/FETCH_CATEGORY_FAILURE';
+/**
+ * Action type dispatched when the fetch categories request starts.
```
catarinacouto, line 38:
```suggestion
 * Action type dispatched when the fetch category request starts.
```

packages/redux/src/categories/actionTypes.ts
```diff
@@ -29,7 +29,21 @@ export const FETCH_CATEGORIES_REQUEST =
...
+ */
+export const FETCH_CATEGORY_REQUEST =
+  '@farfetch/blackout-redux/FETCH_CATEGORY_REQUEST';
+/**
+ * Action type dispatched when the fetch categories request succeeds.
```
catarinacouto, line 43:
```suggestion
 * Action type dispatched when the fetch category request succeeds.
```

catarinacouto (changes requested):

packages/redux/src/categories/actionTypes.ts
```diff
@@ -29,9 +29,23 @@ export const FETCH_CATEGORIES_REQUEST =
...
+ */
+export const FETCH_CATEGORY_REQUEST =
+  '@farfetch/blackout-redux/FETCH_CATEGORY_REQUEST';
+/**
+ * Action type dispatched when the fetch category request succeeds.
+ */
+export const FETCH_CATEGORY_SUCCESS =
+  '@farfetch/blackout-redux/FETCH_CATEGORY_SUCCESS';
+/**
+ * Action type dispatched when the reset category occurs.
```
catarinacouto, line 48:
```suggestion
 * Action type dispatched when the reset categories occur.
```

packages/redux/src/categories/actionTypes.ts
catarinacouto: This one is still missing :)

packages/react/src/categories/hooks/__tests__/useCategories.test.tsx
```diff
@@ -44,126 +38,73 @@ describe('useCategories', () => {
...
+        reset: expect.any(Function),
+      },
     });
   });
 
-  it('should render in categories loading state', () => {
-    const { areCategoriesLoading } = getRenderedHook(
-      mockCategoriesLoadingState,
-    );
+  it('should render loading state', () => {
```
PedroMiguelSS, line 55:
```suggestion
  it('should return the loading state correctly', () => {
```

Please, check the other occurrences as well.

PedroMiguelSS (changes requested):

packages/react/src/categories/hooks/__tests__/useCategory.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+jest.mock('@farfetch/blackout-redux', () => ({
+  ...jest.requireActual('@farfetch/blackout-redux'),
+  resetCategoriesState: jest.fn(() => ({ type: 'reset' })),
+  fetchCategories: jest.fn(() => ({ type: 'fetchCategories' })),
+  fetchCategory: jest.fn(() => ({ type: 'fetchCategory' })),
+}));
+
+const mockDispatch = jest.fn();
+
+jest.mock('react-redux', () => ({
```
andreribeirosilva, line 23: we can avoid the react-redux mock, by mocking the redux factory functions, and hence check the params being passed to the react hook callback (check search intents tests).

Same for all other test files (hooks tests)

packages/react/src/categories/hooks/useCategories.ts
```diff
@@ -1,126 +1,40 @@
-/**
- * Hook to provide data for logic around categories and top categories.
- */
 import {
   areCategoriesFetched as areCategoriesFetchedSelector,
```
andreribeirosilva, line 2: I believe that  these import renames are unnecessary

packages/react/src/categories/hooks/useCategory.ts
```diff
@@ -0,0 +1,51 @@
...
+    isLoading,
+  ]);
+
+  return {
+    isLoading,
+    error,
+    isFetched,
+    data: category,
+    actions: {
+      fetch,
```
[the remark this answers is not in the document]
andreribeirosilva: IMO we should discuss this approach. I believe we have both being applied on our react custom hooks.

packages/react/src/categories/hooks/useTopCategories.ts
```diff
@@ -0,0 +1,40 @@
+import {
+  areTopCategoriesFetched as areTopCategoriesFetchedSelector,
+  areTopCategoriesLoading as areTopCategoriesLoadingSelector,
```
andreribeirosilva, line 3: same

packages/react/src/categories/utils.ts
```diff
@@ -0,0 +1,12 @@
+import type { Category } from '@farfetch/blackout-client';
```
andreribeirosilva, line 1: What do you think about creating a `utils` folder?

catarinacouto (approved):

packages/client/src/categories/getCategory.ts
```diff
@@ -0,0 +1,22 @@
+import { adaptError } from '../helpers/client/formatError';
+import client from '../helpers/client';
+import join from 'proper-url-join';
+import type { GetCategory } from './types';
+
+/**
+ * Method responsible for getting a category by it's id.
```
boliveira, line 7:
```suggestion
 * Method responsible for getting a category by its id.
```

packages/client/src/categories/getCategory.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @param id - Category identifier.
+ * @param config - Custom configurations to send to the client instance (axios).
+ *
+ * @returns Promise that will be resolved when the call to the endpoint finishes.
+ */
+const getCategory: GetCategory = (id, config) =>
```
boliveira, line 14: What do you think of renaming `id` to `categoryId`?

packages/client/src/categories/types/getCategory.types.ts
```diff
@@ -0,0 +1,4 @@
+import type { Category } from './category.types';
+import type { Config } from '../..';
+
+export type GetCategory = (id: number, config?: Config) => Promise<Category>;
```
boliveira, line 4:
```suggestion
export type GetCategory = (id: Category['id'], config?: Config) => Promise<Category>;
```

andreribeirosilva (approved):

packages/react/src/categories/utils/getRootCategory.ts
```diff
@@ -0,0 +1,14 @@
+import type { Category } from '@farfetch/blackout-client';
```
boliveira, line 1: I think this module should be at least inside the `@farfetch/blackout-redux` because if the developer does not want to use the hook and opts to use the redux's actions and selectors, he shouldn't have to add the `@farfetch/blackout-react` dependency for that. Since the `@farfetch/blackout-redux` is a dependency of `@farfetch/blackout-react` both of those use cases benefit from that change. What do you think?

packages/react/src/categories/hooks/useCategory.ts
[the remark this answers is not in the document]
boliveira: Just to add to this, I think Pedro's comment is correct and all hooks that do not follow that approach should be changed to conform to it. The reason is that the hook provides its result from the passed in parameter instead, so if the `fetch` action allows to request a different `id`, the developer will not have access to its result until he changed the `id` parameter of the hook, but if he does that, the `useEffect` will be triggered which will do the fetch anyway (if he has `enableAutoFetch` true of course).

I only think that it should be named `fetch` instead of `refetch` because if the `enableAutoFetch` parameter is false, `refetch` I think does not make much sense.

Another point is that the `fetch` action should be provided by the `useCategories` hook and this hook should depend on that. The `fetch` provided by the `useCategory` hook would only set the `id` parameter to the `id` parameter passed to the hook.

What do you think?

packages/react/src/categories/hooks/useCategory.ts
```diff
@@ -0,0 +1,49 @@
...
+  isCategoryFetched,
+  isCategoryLoading,
+  StoreState,
+} from '@farfetch/blackout-redux';
+import { useAction } from '../../helpers';
+import { useCallback, useEffect } from 'react';
+import { useSelector } from 'react-redux';
+import type { UseCategoryOptions } from './types/useCategory';
+
+const useCategory = (id: number, options: UseCategoryOptions = {}) => {
```
boliveira, line 14: What do you think of this change?

```suggestion
const useCategory = (categoryId: Category['id'], options: UseCategoryOptions = {}) => {
```
I think `categoryId` is a better name for that parameter and the type can be set to `Category['id']`. What do you think?

packages/redux/src/categories/types/action.types.ts
```diff
@@ -28,6 +28,29 @@ export type FetchCategoriesAction =
   | FetchCategoriesSuccessAction
   | FetchCategoriesFailureAction;
 
+export interface FetchCategoryRequestAction extends Action {
+  meta: { id: number };
```
boliveira, line 32:
```suggestion
  meta: { id: Category['id'] };
```

Please, review all locations which define the type of the categoryId value as `number` to be `Category['id']` instead.

packages/redux/src/categories/types/state.types.ts
```diff
@@ -1,6 +1,8 @@
 import type { BlackoutError, Category } from '@farfetch/blackout-client';
 import type { CombinedState } from 'redux';
 
+export type CategoryId = string;
```
boliveira, line 4: Question: Why is this type necessary? There is the `Category['id']` type which is a number and not a string.

packages/react/src/categories/utils/getRootCategory.ts
JoaoFerreira10: Makes sense!!👍

packages/react/src/categories/hooks/useCategory.ts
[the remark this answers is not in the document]
JoaoFerreira10: Done! The `fetch` action to request a different ` id`, was removed! 

Regarding your 2nd point: we have 2 different endpoints: one for the list of categories ('`/commerce/v1/categories`'), and another to fetch a category ('`/commerce/v1/categories/:id`'). So, I think that makes sense to have 2 different actions! =)

PedroMiguelSS (approved):

packages/redux/src/categories/actions/factories/fetchCategoryFactory.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Creates a thunk factory configured with the specified client to fetch a
+ * category by id.
+ *
+ * @param getCategory - Get category client.
+ *
+ * @returns Thunk factory.
+ */
+const fetchCategoryFactory =
+  (getCategory: GetCategory) =>
+  (id: number, config?: Config) =>
```
boliveira, line 23: Missed this one :) 
```suggestion
  (categoryId: Category['id'], config?: Config) =>
```

boliveira (approved):

PR #512, opened 2022-09-27
feat(redux|client|react): add brands hooks

JoaoFerreira10:
## Description

Added brands hooks.

## useBrands

- Params
  - query (optional) 
  - enableAutoFetch (optional)
- Return
  - isLoading
  - error
  - isFetched
  - data
  - actions
    -  fetch
    - reset 

## useBrand

- Params
  - id   
  - enableAutoFetch (optional)
- Return
  - isLoading
  - error
  - isFetched
  - data
  - actions
    -  fetch
    -  refetch

## Checklist

## Disclaimer


packages/react/src/brands/hooks/useBrand.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }, [config, enableAutoFetch, error, fetchBrand, id, isFetched, isLoading]);
+
+  return {
+    isLoading,
+    error,
+    isFetched,
+    data: brand,
+    actions: {
+      fetch,
```
boliveira, line 41: The hook should not allow to fetch a brand id different than the one that it was instantiated with. If the user wants to fetch another brand id, he must change the id parameter for the `useBrand` hook so the useEffect will be re-executed. Is there a reason to provide this that I am not seeing?
If there is not one, then the `refetch` action should be renamed to `fetch` and the current `fetch` should be removed.

packages/react/src/brands/hooks/useBrand.ts
```diff
@@ -0,0 +1,47 @@
...
+  const isFetched = useSelector((state: StoreState) =>
+    isBrandFetched(state, id),
+  );
+  const error = useSelector((state: StoreState) => getBrandError(state, id));
+  const brand = useSelector((state: StoreState) => getBrand(state, id));
+  const fetch = useAction(fetchBrandAction);
+  const fetchBrand = useCallback(() => fetch(id, config), [config, fetch, id]);
+
+  useEffect(() => {
+    if (!isLoading && !error && !isFetched && enableAutoFetch) {
```
boliveira, line 30: I know that this should not happen in TS, but I think it would be a good rule of thumb to also check if the id parameter contains a value. It only should not contain a value if the developer forces a cast to `Brand['id']` though.

Also, you can remove the `!error` condition because it is accounted for by the `!isFetched` condition.

packages/react/src/brands/hooks/useBrands.ts
```diff
@@ -0,0 +1,71 @@
...
+  );
+  const result = useSelector((state: StoreState) =>
+    getBrandsResult(state, hash),
+  );
+  const error = useSelector((state: StoreState) => getBrandsError(state, hash));
+  const fetch = useAction(fetchBrands);
+  const reset = useAction(resetBrandsState);
+
+  useEffect(() => {
+    if (!isLoading && !error && !result && enableAutoFetch) {
```
boliveira, line 44: Instead of `!result` shouldn't it be `!isFetched`?

packages/react/src/brands/hooks/useBrand.ts
```diff
@@ -0,0 +1,47 @@
...
+  isBrandLoading,
+  StoreState,
+} from '@farfetch/blackout-redux';
+import { useAction } from '../../helpers';
+import { useCallback, useEffect } from 'react';
+import { useSelector } from 'react-redux';
+import type { Brand } from '@farfetch/blackout-client';
+import type { UseBrandOptions } from './types/useBrand';
+
+const useBrand = (id: Brand['id'], options: UseBrandOptions) => {
```
boliveira, line 15: I think if it was named `brandId` would be easier to understand. What do you think?

JoaoFerreira10: I agree too!

packages/react/src/brands/hooks/useBrand.ts
JoaoFerreira10: Yeah, I noticed that `!error `and `!isLoading` it's accounted for by the `!isFetched` condition!  
I only added these 2 conditions to keep the same approach that we had applied for the other hooks.  

I can remove these conditions, but  I would suggest creating a new ticket to refactor the other hooks.

packages/react/src/brands/hooks/useBrand.ts
JoaoFerreira10: Done! And I applied the same approach to the `useCategory` hook too! 👍

boliveira (approved):

PR #519, opened 2022-10-03
feat(analytics|react): share uniqueEventId between analytics' integrations

ruifcnunes:
## Description
This PR adds a new analytics event context property called `__uniqueEventId` that will identify a single event throughout all integrations, so we can easily correlate an event across them. 

For now, only Omnitracking, GA4 and GTM will make use of this new property, but since it is calculated on the analytics layer, we can later use it on the remaining integrations as the business sees fit.

## Checklist

## Disclaimer


danielbento92 (approved):

packages/react/src/analytics/integrations/GA4/types/index.ts
```diff
@@ -48,7 +48,10 @@ export type OnPreProcessCommandsHandler = (
 ) => GA4CommandList;
 
 export type GA4CommandList = GA4Command[];
-export type GA4Command = [command: string, ...args: unknown[]];
+export type GA4Command = [
+  command: string,
+  ...args: unknown[] | Record<string, string>[],
```
boliveira, line 53: I think we should continue use the unknown here as it is not certain that all commands receive a `Record<string, string>`. Ideally, we could type this differently for the `event` and `config` commands but I believe for this case we can let it as is. I will suggest another change where this is being used in the code.

packages/react/src/analytics/integrations/GA4/GA4.ts
```diff
@@ -377,11 +380,39 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
...
+   */
+  attachUniqueEventIdToCommand(
+    command: GA4Command,
+    data:
+      | TrackEventData
+      | PageviewEventData
+      | LoadIntegrationEventData
+      | SetUserEventData,
+  ) {
+    const eventProperties = command[2] as Record<string, string>;
```
boliveira, line 406: This could be:

```
   const eventProperties = command[2];

    if (eventProperties && typeof eventProperties === 'object') {
      const uniqueEventId = get(
        data,
        'context.event.__uniqueEventId',
      ) as string;

      (eventProperties as Record<string, string>)[OPTION_UNIQUE_EVENT_ID_KEY] =
        uniqueEventId;
    }

    return command;
```

As there is a check already to see if  `eventProperties` is an object, we can take advantage of that and with a cast that is valid inside the if, we can make this type safe.

packages/react/src/analytics/integrations/GA4/GA4.ts
```diff
@@ -377,11 +380,39 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
...
+    const uniqueEventId = get(data, 'context.event.__uniqueEventId');
```
boliveira, line 407: Also, we could extract the `__uniqueEventId` to a constant and replace it in all integrations. That way we can easily change the name of the `__uniqueEventId` property in a single point.

boliveira (approved):

PR #524, opened 2022-10-07
feat(analytics|react): add product details and listing pages to omnitracking

danielbento92:
## Description

- Add persist of navigation from parameter, by using from parameter on page events;
- Add Product Listing Page Event to omnitracking;
- Add Product Page Event to omnitracking;
- Since document.referrer stays the same on single page applications,
this PR adds an alternative property called `pageLocationReferrer`
for page trackings that will hold the previous URL when navigating trough
pages.
- This PR also adds the type of event that is being tracked to each `useContext`
function and calls the webContext directly on the analyticsWeb constructor.
Mapped the Omnitracking's `referrer` to the new `pageLocationReferrer`,
since the document.referrer was causing issues when analyzing user behavior.
- Adjust test to cover the above points.

## Checklist

## Disclaimer

### DIFF
--- a/packages/analytics/src/Analytics.ts
+++ b/packages/analytics/src/Analytics.ts
@@ -18,6 +18,7 @@ import type {
   EventContextData,
   EventData,
   EventProperties,
+  ExtendedTrackTypes,
   IntegrationFactory,
   IntegrationOptions,
   IntegrationRuntimeData,
@@ -663,12 +664,7 @@ class Analytics {
    *
    * @returns Data for the event.
    */
-  protected async getEventData<
-    T extends
-      | TrackTypesValues
-      | typeof ON_SET_USER_TRACK_TYPE
-      | typeof LOAD_INTEGRATION_TRACK_TYPE,
-  >(
+  protected async getEventData<T extends ExtendedTrackTypes>(
     type: T,
     additionalData?: Partial<EventData<T>>,
     eventContext?: EventData<T>['context']['event'],
--- a/packages/analytics/src/integrations/Omnitracking/__tests__/Omnitracking.test.ts
+++ b/packages/analytics/src/integrations/Omnitracking/__tests__/Omnitracking.test.ts
@@ -114,11 +114,11 @@ describe('Omnitracking', () => {
 
       expect(postTrackingSpy).toHaveBeenCalledWith({
         ...mockExpectedPagePayloadWeb,
-        parameters: {
+        parameters: expect.objectContaining({
           ...mockExpectedPagePayloadWeb.parameters,
           uniqueViewId: expect.any(String),
           previousUniqueViewId: null,
-        },
+        }),
       });
     });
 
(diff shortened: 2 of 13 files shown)


packages/analytics/src/integrations/Omnitracking/__tests__/Omnitracking.test.ts
```diff
@@ -719,13 +719,34 @@ describe('Omnitracking', () => {
...
-            )(mockedTrackData),
+            )(trackEventsData[eventMapperKeyTemp]),
           ).toMatchSnapshot();
           expect(typeof definitions.trackEventsMapper[eventMapperKeyTemp]).toBe(
             'function',
           );
         },
       );
+
+      fit.each(Object.keys(definitions.pageEventsMapper))(
```
boliveira, line 730: Lingering fit

packages/analytics/src/Analytics.ts
```diff
@@ -113,17 +114,22 @@ class Analytics {
...
+    key: string | undefined,
+    eventType?: ExtendedTrackTypes,
+  ): Promise<unknown>;
+  async context(key?: string, eventType?: ExtendedTrackTypes) {
     const externalContextData = {};
 
     for (const contextFn of this.contextFns) {
       try {
-        const contextFnResult = await contextFn();
+        const contextFnResult = await contextFn(eventType);
```
boliveira, line 132: Question: Why is it necessary to support different contexts by event type? Is there a specific case where you need to provide different contexts for different event types?

packages/react/src/analytics/context.ts
```diff
@@ -12,26 +13,46 @@ export type WebContextType = {
...
  */
-export const context = (): WebContextType => {
+export const context = (
+  trackEventType?: ExtendedTrackTypes,
+): WebContextType => {
+  const pageLocationReferrer = lastLocation;
+  const locationHref = window.location.href;
+  const isPageEvent = trackEventType === trackTypes.PAGE;
+
+  if (isPageEvent && lastLocation !== locationHref) {
```
boliveira, line 37: Is there a case where the location changes but the event type is not page? What I mean is, why do we need to include the `isPageEvent` for this calculation? Wouldn't this be enough?

```suggestion
  if (lastLocation !== locationHref) {
```

ruifcnunes (approved):

packages/react/src/analytics/context.ts
danielbento92: We must ensure this variable is only updated when a page view occurs, otherwise we would end up with page action events being sent with the incorrect location referrer, as this variable would be updated to the current page URL instead of holding the previous one until a new page view gets triggered.

packages/analytics/src/integrations/Omnitracking/__tests__/Omnitracking.test.ts
danielbento92: Solved

packages/react/src/analytics/context.ts
boliveira: I understand the problem but I think that this should not be a concern for our context / package to have. If there is an event that is tracked in a new page before the page event is tracked, it would mean the context would register the `pageLocationReferrer` with an incorrect url. I think the context should not concern with that and it should be the responsibility of the application to make sure that a page event is triggered before an event is triggered when a new page is rendered. What do you think?

danielbento92: Updated: Removed dependency of eventType from context.

boliveira (approved):

PR #535, opened 2022-10-13
feat(analytics): add omnitracking's wishlist page tracking mappings

danielbento92:
## Description

- Added wishlist event to omnitracking mappper.

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/analytics/src/integrations/Omnitracking/__tests__/__snapshots__/Omnitracking.test.ts.snap
+++ b/packages/analytics/src/integrations/Omnitracking/__tests__/__snapshots__/Omnitracking.test.ts.snap
@@ -81,6 +81,14 @@ Object {
 }
 `;
 
+exports[`Omnitracking track events definitions \`Share\` return should match the snapshot 1`] = `
+Object {
+  "actionArea": "PDP",
+  "productId": "123456",
+  "tid": 1205,
+}
+`;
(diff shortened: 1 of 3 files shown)


packages/analytics/src/integrations/Omnitracking/__tests__/omnitracking-helper.test.ts
```diff
@@ -101,3 +102,21 @@ describe('getCheckoutEventGenericProperties', () => {
     });
   });
 });
+
+describe('getProductLineItemsQuantity', () => {
+  it('should return quantity equals to 10', () => {
```
boliveira, line 107: Nitpicking: When describing the test, describe what the logic should be instead of being too specific, something like this:

```suggestion
  it('should return the sum of all quantities from all products in the list', () => {
```

Both ways describe the same thing but the first one is more generic and I can understand that what it is doing is summing all the quantities more easily.

boliveira (approved):

packages/analytics/src/integrations/Omnitracking/__tests__/omnitracking-helper.test.ts
danielbento92: Fixed

PR #539, opened 2022-10-13
feat(react|redux|client): refactor and add seo metadata hooks and utils

franciscofarinha:
## Description

- Refactor `useMetaTags` into `useSeoMetadata`
- add `useListingSeoMetadataParams`
- add `getProductSeoMetadataParams`
- add `flattenObject`
- add metadata to contents `serverInitialState`

## Checklist

## Disclaimer


packages/client/src/utils/flattenObject.ts
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+type NotObject<T> = Exclude<T, Record<string, any>>;
+type FlattenedObject<T> = Record<string, NotObject<T>>;
+
+export default function flattenObject<T>(
```
boliveira, line 5: Generic functions should use the type parameters more than once to relate values. In this case it is being used only in one place, so it is not providing value for the consumers. I believe the correct function definition should be quite a bit more involved (like defining another type on the fly that will add recursively all properties that are of type object to the passed type).

Let's simplify this a bit and remove the generic definition and return a Record<string, unknown>. Also the `object` parameter could be renamed to sourceObj or just source and could be of type Record<string, unknown> as well.

packages/react/src/contents/hooks/useSeoMetadata.ts
```diff
@@ -0,0 +1,92 @@
...
+ * @param data - object that receives SEO query, icon links, additional links, metas and options
+ *
+ * @returns - Returns actions and SEO metadata state for a given path.
+ */
+const useSeoMetadata = ({
+  query,
+  appIconLinks = getDefaultAppLinks(),
+  links = [],
+  metas = [],
+  options = {},
```
boliveira, line 38: To be consistent with the other hooks design, can you move the options `property` to the second parameter?

packages/react/src/contents/hooks/__tests__/__fixtures__/useListingSeoMetadataParams.fixtures.ts
```diff
@@ -0,0 +1,134 @@
+export const mockFilterSegments = {
```
boliveira, line 1: Can you rename this file to be `getSeoMetadataParams.fixtures.ts`?

packages/react/src/contents/hooks/useSeoMetadata.ts
```diff
@@ -0,0 +1,93 @@
...
+  const isFetched = useSelector((state: StoreState) =>
+    isSEOMetadataFetched(state, query),
+  );
+  const seo = useSelector((state: StoreState) =>
+    getSEOMetadataResult(state, query),
+  );
+  const { path = '' } = query;
+  const fetch = useAction(fetchSEOMetadata);
+
+  const data = seo
```
andreribeirosilva, line 57: Can we memoize this object?

andreribeirosilva (approved):

PR #543, opened 2022-10-17
feat(redux): configurations actions

giuliamml:
## Description
This PR adds the redux logic for `fetchConfiguration` and `fetchConfigurations` actions, reducers and selectors.

## Checklist

## Disclaimer


packages/redux/src/settings/actions/factories/fetchConfigurationsFactory.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Creates a thunk factory configured with the specified client to fetch all
+ * configurations.
+ *
+ * @param getConfigurations - Get configurations client.
+ *
+ * @returns Thunk factory.
+ */
+const fetchConfigurationsFactory =
+  (getConfigurations: GetConfigurations) =>
+  (config?: Config, query?: ConfigurationsQuery) =>
```
boliveira, line 24: Same as https://github.com/Farfetch/blackout/pull/543/files#r997069069

packages/redux/src/settings/selectors/configuration.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @param state - Application state.
+ *
+ * @returns Error information (`null` if there are no errors).
+ */
+export const getConfigurationError = (
+  state: StoreState,
+  configurationCode: Configuration['code'],
+) =>
+  getError(state.settings?.configurations.configuration as ConfigurationState)[
```
boliveira, line 18: Instead, change it to this, please: 

```suggestion
  getError((state.settings as SettingsState).configurations.configuration)[
    configurationCode
  ];
```

This is to intentionally force an error if the `settings` reducer is not configured on the redux store. 
Please, check all other selectors and change them to work like this.

packages/redux/src/settings/types/action.types.ts
```diff
@@ -0,0 +1,62 @@
...
+  | FetchConfigurationsFailureAction;
+
+export interface FetchConfigurationRequestAction extends Action {
+  meta: { code: Configuration['code'] };
+  type: typeof actionTypes.FETCH_CONFIGURATION_REQUEST;
+}
+export interface FetchConfigurationSuccessAction extends Action {
+  meta: { code: Configuration['code'] };
+  type: typeof actionTypes.FETCH_CONFIGURATION_SUCCESS;
+  payload: Payload;
```
boliveira, line 42: I believe the Payload type is not correct here as the result property is different. Here it will receive a single Configuration object instead of an array. I think it is better if you create another type just for the FetchConfigurationSuccessAction instead of reusing the same type used by the FetchConfigurationsSuccessAction.

packages/redux/src/settings/index.ts
```diff
@@ -0,0 +1,8 @@
+export * as configurationsActionTypes from './actionTypes';
```
boliveira, line 1:
```suggestion
export * as settingsActionTypes from './actionTypes';
```

packages/redux/src/settings/reducer/index.ts
```diff
@@ -0,0 +1,7 @@
+import configurationsReducer, { entitiesMapper } from './configurations';
+
+export const entitiesMapperConfigurations = {
```
boliveira, line 3: Can you rename it to that, please? This is to keep consistency.

```suggestion
export const configurationsEntitiesMapper = {
```

packages/client/src/settings/types/configuration.ts
```diff
@@ -1,27 +1,32 @@
...
+  scopes: string[];
+};
+
+export type ConfigurationProperty = {
+  code: string;
+  description: string;
+  value: string;
+  schemaFieldType: ConfigurationSchemaFieldType;
+  security: ConfigurationSecurity;
+  contextValues?: [{ code: string; value: string }];
```
jhairabreu, line 22: Can you change this to  `Array<{ code: string; value: string }>` or `{ code: string; value: string }[]` , for consistency? It's how we are doing in the project.

jhairabreu (changes requested):

packages/client/src/settings/types/configuration.ts
giuliamml: changed to `Array<{ code: string; value: string }>`

boliveira (approved):

jhairabreu (approved):

PR #550, opened 2022-10-19
chore(react|redux): remove checkout hooks

boliveira:
## Description

- This removes `useCheckout` hook temporarily in order to not introduce
a breaking change when the next version of the package is released.
The redux implementation of both checkout and payments
areas were checked and tweaked a little bit by fixing reducers
and adding missing reset state actions.
- Renamed `getPaymentMethods` and `fetchPaymentMethods` to
`getCheckoutOrderPaymentMethods` and `fetchCheckoutOrderPaymentMethods`
respectively and moved those modules to the checkout area from payments.
- Changed `renderContent` function signature to accept a boolean
parameter instead of a string.
- Changed reducerFactory/createReducerWithResult exports to remove the
`validateReset` function call
and allow to pass action types that should reset
the reducer's state.

BREAKING CHANGE:
- `useCheckout` hook was removed from @farfetch/blackout-react package.
- `getCheckoutOrderOperation` and `fetchCheckoutOrderOperation` signatures
have changed to accept separate parameters for the checkoutOrderId and
operationId values instead of using an object.
- `getPaymentMethods` client was renamed to `getCheckoutOrderPaymentMethods`.
- `fetchPaymentMethods` action was renamed to `fetchCheckoutOrderPaymentMethods`.

## Checklist

## Disclaimer


jhairabreu (approved):

giuliamml (approved):

catarinacouto (approved):

packages/redux/src/returns/selectors.ts
```diff
@@ -165,3 +166,27 @@ export const areReturnPickupCapabilitiesLoading = (state: StoreState) =>
...
+/**
+ * Returns the loading status for the get reschedule request.
+ *
+ * @function
+ *
+ * @param {object} state - Application state.
+ *
+ * @returns {boolean} Loading status.
+ */
+export const isPickupRescheduleRequestsLoading = (state: StoreState) =>
```
boliveira, line 179: Can you change it to `arePickupRescheduleRequestsLoading`?

packages/redux/src/returns/__tests__/reducer.test.ts
```diff
@@ -68,6 +68,9 @@ describe('returns reducer', () => {
       actionTypes.FETCH_RETURN_PICKUP_CAPABILITIES_REQUEST,
       actionTypes.FETCH_RETURN_REQUEST,
       actionTypes.UPDATE_RETURN_REQUEST,
+      actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUEST_REQUEST,
+      actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUESTS_REQUEST,
+      actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUEST_REQUEST,
```
boliveira, line 73: Duplicated action type.

packages/redux/src/returns/__tests__/reducer.test.ts
```diff
@@ -249,6 +285,33 @@ describe('returns reducer', () => {
       ).toBe(initialState.isLoading);
     });
 
+    it('should handle FETCH_RETURN_PICKUP_RESCHEDULE_REQUEST_FAILURE action type', () => {
+      expect(
+        reducer(undefined, {
```
boliveira, line 290: For these tests, it is better to define a state where `isLoading` is different than `initialState.isLoading` to make sure the test will fail when the code does.

packages/redux/src/returns/reducer.ts
```diff
@@ -30,11 +34,17 @@ const error = (state = INITIAL_STATE.error, action: AnyAction) => {
...
+    case actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUESTS_FAILURE:
+    case actionTypes.CREATE_RETURN_PICKUP_RESCHEDULE_REQUEST_FAILURE:
       return action.payload.error;
     case actionTypes.CREATE_RETURN_REQUEST:
     case actionTypes.FETCH_RETURN_PICKUP_CAPABILITIES_REQUEST:
     case actionTypes.FETCH_RETURN_REQUEST:
     case actionTypes.UPDATE_RETURN_REQUEST:
+    case actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUEST_REQUEST:
+    case actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUESTS_REQUEST:
+    case actionTypes.CREATE_RETURN_PICKUP_RESCHEDULE_REQUEST_REQUEST:
```
boliveira, line 47: There is a bug here that I fixed in another PR but you may include as well. This is missing a return statement to clear the error for these action types:
`return INITIAL_STATE.error`

packages/redux/src/returns/reducer.ts
```diff
@@ -65,6 +78,12 @@ const isLoading = (state = INITIAL_STATE.isLoading, action: AnyAction) => {
     case actionTypes.FETCH_RETURN_FAILURE:
     case actionTypes.UPDATE_RETURN_SUCCESS:
     case actionTypes.UPDATE_RETURN_FAILURE:
+    case actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUEST_SUCCESS:
+    case actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUEST_FAILURE:
+    case actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUESTS_SUCCESS:
+    case actionTypes.FETCH_RETURN_PICKUP_RESCHEDULE_REQUESTS_FAILURE:
+    case actionTypes.CREATE_RETURN_PICKUP_RESCHEDULE_REQUEST_SUCCESS:
+    case actionTypes.CREATE_RETURN_PICKUP_RESCHEDULE_REQUEST_FAILURE:
```
boliveira, line 86: Same bug, missing `return INITIAL_STATE.isLoading` here

packages/redux/src/returns/reducer.ts
```diff
@@ -113,6 +132,16 @@ const pickupCapabilities = reducerFactory(
   actionTypes,
 );
 
+const pickupRescheduleRequests = reducerFactory(
+  [
+    'FETCH_RETURN_PICKUP_RESCHEDULE_REQUEST',
+    'FETCH_RETURN_PICKUP_RESCHEDULE_REQUESTS',
```
boliveira, line 138: Is this ok to mix the `FETCH_RETURN_PICKUP_RESCHEDULE_REQUEST` with `FETCH_RETURN_PICKUP_RESCHEDULE_REQUESTS` as they have a different payload?

boliveira: Actually, the payload is being ignored and because of that there is no selector for it. Is this correct? Shouldn't this be stored somewhere on the store and a selector be available to fetch it from there?

packages/redux/src/returns/selectors.ts
andreecosta: done!

packages/redux/src/returns/__tests__/reducer.test.ts
andreecosta: fixed.

packages/redux/src/returns/__tests__/reducer.test.ts
andreecosta: done!

packages/redux/src/returns/reducer.ts
andreecosta: done!

packages/redux/src/returns/reducer.ts
andreecosta: done!

packages/redux/src/returns/reducer.ts
andreecosta: Just leaving the note: We had a meet about this. This is something that needs to be decided the best way to be done (not having the results).

PR #556, opened 2022-10-21
feat(react): settings hooks

giuliamml:
This PR adds the following hooks to the settings folder: `useConfiguration` and `useConfigurations`

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/settings/hooks/__tests__/useConfiguration.test.tsx
+++ b/packages/react/src/settings/hooks/__tests__/useConfiguration.test.tsx
@@ -1,0 +1,127 @@
+import { cleanup, renderHook } from '@testing-library/react';
+import { fetchConfiguration } from '@farfetch/blackout-redux';
+import {
+  mockConfigurationCode,
+  mockConfigurationErrorState,
+  mockConfigurationLoadingState,
+  mockConfigurationsInitialState,
+  mockConfigurationsState,
+} from 'tests/__fixtures__/settings';
+import { useConfiguration } from '../..';
+import { withStore } from '../../../../tests/helpers';
+
+jest.mock('@farfetch/blackout-redux', () => ({
+  ...jest.requireActual('@farfetch/blackout-redux'),
+  fetchConfiguration: jest.fn(() => () => Promise.resolve()),
+}));
+
+const getRenderedHook = (
+  state = mockConfigurationsInitialState,
+  config = {},
+) => {
+  const {
+    result: { current },
+  } = renderHook(() => useConfiguration(mockConfigurationCode, config), {
+    wrapper: withStore(state),
+  });
+
+  return current;
+};
+
+describe('useConfiguration', () => {
+  beforeEach(jest.clearAllMocks);
+  afterEach(cleanup);
+
+  it('should return values correctly with initial state', () => {
+    const current = getRenderedHook(mockConfigurationsState);
+
+    expect(current).toStrictEqual({
+      isFetched: true,
+      isLoading: false,
+      data: mockConfigurationsState.entities.configurations[
+        mockConfigurationCode
+      ],
+      error: null,
+      actions: {
+        fetch: expect.any(Function),
+      },
+    });
+  });
+
+  it('should return the loading state correctly', () => {
+    const data = getRenderedHook(mockConfigurationLoadingState);
+
+    expect(data.isLoading).toBe(
+      mockConfigurationLoadingState.settings.configurations.configuration
+        .isLoading[mockConfigurationCode],
+    );
+  });
+
+  it('should return the error state correctly', () => {
+    const data = getRenderedHook(mockConfigurationErrorState);
+
+    expect(data.error).toEqual(
+      mockConfigurationErrorState.settings.configurations.configuration.error[
+        mockConfigurationCode
+      ],
+    );
+  });
+
+  it('should return the fetched state correctly', () => {
+    const { isFetched } = getRenderedHook(mockConfigurationsState);
+
+    expect(isFetched).toBe(true);
+  });
+
+  describe('options', () => {
+    it('should call `fetch` action if `enableAutoFetch` option is true', () => {
+      getRenderedHook();
+
+      expect(fetchConfiguration).toHaveBeenCalledWith(
+        mockConfigurationCode,
+        undefined,
+        undefined,
+      );
+    });
+
+    it('should not call `fetch` action if `enableAutoFetch` option is false', () => {
+      getRenderedHook(mockConfigurationsState, {
+        enableAutoFetch: false,
+      });
+
+      expect(fetchConfiguration).not.toHaveBeenCalled();
+    });
+  });
+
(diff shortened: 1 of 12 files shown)


packages/react/src/settings/hooks/useConfiguration.ts
```diff
@@ -0,0 +1,57 @@
...
+  );
+  const fetch = useAction(fetchConfigurationAction);
+
+  const fetchConfiguration = useCallback(
+    () => fetch(configurationCode, fetchQuery, fetchConfig),
+    [fetch, configurationCode, fetchQuery, fetchConfig],
+  );
+
+  useEffect(() => {
+    if (!isLoading && !error && !isFetched && enableAutoFetch) {
```
boliveira, line 41:
```suggestion
    if (!isLoading && !isFetched && enableAutoFetch && configurationCode) {
```

tests/__fixtures__/settings/configurations.fixtures.ts
```diff
@@ -35,15 +41,16 @@ export const mockConfigurationsInitialState = {
     configurations: {
       error: null,
       isLoading: false,
+      isFetched: false,
```
boliveira, line 44: `isFetched` is not a part of the reducer state

packages/react/src/settings/hooks/__tests__/useConfigurations.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  describe('options', () => {
+    it('should call `fetch` action if `enableAutoFetch` option is true', () => {
+      getRenderedHook();
+
+      expect(fetchConfigurations).toHaveBeenCalled();
+    });
+
+    it('should not call `fetch` action if `enableAutoFetch` option is false', () => {
+      getRenderedHook(mockConfigurationsState, {
```
boliveira, line 84: This test needs to use `mockConfigurationsInitialState` because the `mockConfigurationsState` already contains configurations data, so the hook would not fetch even if `enableAutoFetch` was true.

packages/react/src/settings/hooks/useConfiguration.ts
giuliamml: I agree that `configurationCode` should be included as condition (as is a required required param of `fechConfiguration()` ). Shouldn't `!error` also be a condition?

packages/react/src/settings/hooks/useConfigurations.ts
```diff
@@ -0,0 +1,52 @@
...
+    fetch,
+    fetchQuery,
+    fetchConfig,
+  ]);
+
+  return {
+    isLoading,
+    error,
+    isFetched,
+    data: useMemo(
```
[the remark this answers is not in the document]
giuliamml: Actually yes, more logical to return the configurations object. I believe that using the useMemo hook is not really necessary after all

packages/react/src/settings/hooks/__tests__/useConfigurations.test.tsx
giuliamml: replaced to the correct mocked data. Thanks

packages/react/src/settings/hooks/useConfiguration.ts
boliveira: It is not necessary because for `isFetched` to be true one of 2 conditions need to be true:
 - It does have an error or it does have data.
 - It is not loading.
So when you include `!isFetched` in the if statement, you will only need to check that it is not loading to make sure that when `isFetched` is false it is due to not having an error or not having data. I am aware that other hooks still check for it explicitly (which does not hurt but is unnecessary). I will normalize them in a future PR.
Make sense?

giuliamml: right, now it makes sense. `!isFetched` already covers the `!error` condition, I now see why it is not necessary. Thanks

packages/react/src/settings/hooks/useConfiguration.ts
```diff
@@ -0,0 +1,63 @@
...
+  const isFetched = useSelector((state: StoreState) =>
+    isConfigurationFetched(state, configurationCode),
+  );
+  const configuration = useSelector((state: StoreState) =>
+    getConfiguration(state, configurationCode),
+  );
+  const fetch = useAction(fetchConfigurationAction);
+
+  const fetchConfiguration = useCallback(
+    () => fetch(configurationCode, fetchQuery, fetchConfig),
```
boliveira, line 36: I missed this: Can you add a check for `configurationCode` to make sure it is defined before executing fetch?
Something like this:

```suggestion
    () => {
      if(!configurationCode) {
        return Promise.reject(new Error('Invalid parameter `configurationCode` for `fetch`'));
       }
       
       return fetch(configurationCode, fetchQuery, fetchConfig);
    }
```

Add a corresponding test as well please when `useConfiguration` is called without a configurationCode.

andreribeirosilva (approved):

tests/__fixtures__/settings/configurations.fixtures.ts
```diff
@@ -104,29 +105,28 @@ export const mockConfigurationLoadingState = {
...
   },
 };
 
 export const mockConfigurationsErrorState = {
   settings: {
     configurations: {
-      error: {
-        message: 'An awesome, fascinating and incredible error',
-      },
+      error: null,
```
jhairabreu, line 119: Since this a mock for the error state, shouldn't you have the error instead of `null`?

jhairabreu (changes requested):
Other than my comment, looks good to me

tests/__fixtures__/settings/configurations.fixtures.ts
giuliamml: you are right. I've updated it to the correct mock error state. Thanks

jhairabreu (approved):

packages/react/src/settings/hooks/__tests__/useConfiguration.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+  mockConfigurationLoadingState,
+  mockConfigurationsInitialState,
+  mockConfigurationsState,
+} from 'tests/__fixtures__/settings';
+import { useConfiguration } from '../..';
+import { withStore } from '../../../../tests/helpers';
+
+jest.mock('@farfetch/blackout-redux', () => ({
+  ...jest.requireActual('@farfetch/blackout-redux'),
+  fetchConfigurations: jest.fn(() => () => Promise.resolve()),
```
catarinacouto, line 15: This mock is needed here? I believe not because in not used on the hook file :) What do you think?

catarinacouto (approved):

packages/react/src/settings/hooks/__tests__/useConfiguration.test.tsx
giuliamml: yes, correct! I'v removed it :) Thanks

PR #558, opened 2022-10-24
feat(react): sort recently viewed products

JoaoFerreira10:
## Description

Sort recently viewed products by visualization date.

## Checklist

## Disclaimer


packages/react/src/products/hooks/useRecentlyViewedProducts.ts
```diff
@@ -51,13 +51,20 @@ const useRecentlyViewedProducts = (
     isLoading: isProductListLoading,
     isFetched: isProductListFetched,
     error: producListError,
-    data: { items } = {},
+    data: { items = [] } = {},
```
boliveira, line 54: I think it is preferable to not have a default value for product listing response and instead to make the check for the items when you need to sort the items array. This is because the `products` property in `data` should be undefined while it is loading data or there is an error.

andreribeirosilva (approved):

PR #575, opened 2022-11-04
feat(react|redux|client): add content hooks

andreribeirosilva:
## Description

This PR refactors the following blackout-react hooks:

## useContentType

- Params
    - contentTypeCode
    - options
        - enableAutoFetch (optional)
        - spaceCode
        - environmentCode
        - codes
        - target
        - sort
        - page
        - pageSize
        - metadataSearchTagsValues
        - metadataCustom
- Return
    - isLoading
    - error
    - isFetched
    - data
        - entries
        - pagination
    - actions
        - refetch

## useCommercePages

- Params
    - type
    - id
    - gender
    - brand
    - category
    - priceType
    - sku
    - pageIndex
    - pageSize
- Return
    - isLoading
    - error
    - isFetched
    - data
        - entries
        - pagination
    - actions
        - refetch

Also, in this PR, the commerce pages endpoint is updated and the `useWidget`, `usePage` and `useNavbar` hooks are removed

## Checklist

## Disclaimer


packages/react/src/contents/hooks/types/useCommercePages.types.ts
```diff
@@ -0,0 +1,11 @@
+import type { Config, QueryCommercePages } from '@farfetch/blackout-client';
+
+export interface UseCommercePageOptions extends QueryCommercePages {
+  enableAutoFetch?: boolean;
+  strategy?: string;
+  fetchConfig?: Config;
+}
+
+export enum Strategy {
```
boliveira, line 9: I think it would be best to rename this to something like `CommercePagesStrategy` as `Strategy` is too general. What do you think?

packages/react/src/contents/hooks/types/useCommercePages.types.ts
```diff
@@ -0,0 +1,11 @@
+import type { Config, QueryCommercePages } from '@farfetch/blackout-client';
+
+export interface UseCommercePageOptions extends QueryCommercePages {
```
boliveira, line 3:
```suggestion
export interface UseCommercePagesOptions extends QueryCommercePages {
```

packages/react/src/contents/hooks/useCommercePages.ts
```diff
@@ -1,91 +1,87 @@
...
-    /**
-     * Error state for a specific Commerce Page.
-     */
+    isFetched,
     error,
-    /**
-     * Fetch Commerce Page content for a specific slug.
-     */
-    fetchCommercePages,
+    data: {
```
boliveira, line 78: If `entries` is undefined, `data` will contain an object with all values undefined. I think it would be best if you changed the data calculation to a `useMemo` where you would check if `entries` is undefined, you would return `undefined`. Another question: Can `entries` be undefined but `pagination` not be undefined? If that is not possible, then I would change the `data` calculation to be something like this:

```
const data = useMemo(() => {
  if(!entries || !contentGroup) {
     return undefined;
  }

  return {
   entries:
   pagination: {
      pageIndex: contentGroup?.number,
      totalPages: contentGroup?.totalPages,
      totalItems: contentGroup?.totalItems,
   }

}, [entries, contentGroup]);
```

packages/react/src/contents/hooks/useCommercePages.ts
```diff
@@ -1,91 +1,87 @@
...
+      entries,
+      pagination,
+    },
+    actions: {
+      refetch: fetch,
```
boliveira, line 83: I think this should be called `fetch` because a `refetch` would only be ok if it was guaranteed that the hook would always make a fetch, which would not be the case if `enableAutoFetch` is false.

packages/react/src/contents/hooks/useCommercePages.ts
```diff
@@ -1,91 +1,87 @@
...
   );
-  const action = useAction(fetchCommercePagesAction);
-  const fetchCommercePages = () => action(queryCommerce, slug, strategy);
-  const resetContents = useAction(resetContentsAction);
 
   useEffect(() => {
-    !commercePage && fetchCommercePages();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [commercePage, slug]);
+    if (!isLoading && !error && !isFetched && enableAutoFetch) {
```
boliveira, line 69: You only need this:

```suggestion
    if (!isLoading && !isFetched && enableAutoFetch) {
```

`isFetched` will always be false when there is no error or no data but it will also be false when `isLoading` is true, so we need to check for `isLoading` as well to exclude that case.

packages/react/src/contents/hooks/useContentType.ts
```diff
@@ -1,86 +1,82 @@
 import {
```
boliveira, line 1: The comments regarding `data` calculation, `refetch` action and the useEffect condition for the auto fetch that were raised for the `useCommercePages` hook also apply for this hook as well. To avoid repetition, can you please take a look at those points and change here as well?

packages/react/src/contents/hooks/useCommercePages.ts
```diff
@@ -1,91 +1,87 @@
...
-  const queryContent = {
-    codes: slug,
-    contentTypeCode: 'commerce_pages',
-    pageIndex,
-    pageSize,
-  };
+  const {
+    enableAutoFetch = true,
+    fetchConfig,
+    strategy = Strategy.Default,
```
boliveira, line 26: From what I can see after analysing the code from both the `fetchCommercePagesFactory` action and `getCommercePages` client, it seems to me that if `strategy` is not passed, everything works. If that is the case, do we really need to apply a default value here for strategy?

packages/redux/src/contents/selectors.ts
```diff
@@ -282,5 +317,5 @@ export const getSEOMetadataResult = (
  *
  * @returns Content normalized.
  */
-export const getContent = (state: StoreState, hash: string) =>
-  getEntityById(state, 'contents', hash);
+export const getContent = <T>(state: StoreState, hash: string) =>
+  getEntityById(state, 'contents', hash) as unknown as ContentEntry<T>;
```
boliveira, line 321: `getEntityById` will return undefined if `hash` is not found, so the cast to unknown should not be made. Can you please remove it and reapply the case to `ContentEntry[]` in the getContents selector?

boliveira (approved):

PR #599, opened 2022-12-02
fix(analytics): reset analytics' setUser promise when user is anonymised - Next

ruifcnunes:
## Description
This PR fixes an issue on the analytics flow when the user logs out (gets anonymised), the setUser promise must be reseted so all following track requests await for the new user data to be loaded on the website (and consequently on the analytics package).

## Checklist

## Disclaimer


packages/analytics/src/Analytics.ts
```diff
@@ -406,7 +406,11 @@ class Analytics {
     try {
       this.userInstance && this.userInstance.anonymize();
 
-      this.handleOnUserChanged();
+      await this.handleOnUserChanged();
+
+      this.setUserPromise = new Promise(resolve => {
```
boliveira, line 411: I think this line should be set before calling `handleOnUserChanged` method because of the await it might take a while to come back and execute the change of the `setUserPromise` and while that is going on, another track can happen.

danielbento92 (approved):

PR #601, opened 2022-12-02
fix(react|redux): fix wishlist and bag selectors

boliveira:
## Description

- This fixes `getWishlistItems` selector that was recalculating its value unnecessarily every time the state changed even though it was using `createSelector` because of an incorrect function input to createSelector that was selecting the whole state.
- Changed `getBagItemAvailableSizes` selector to not use `getBagItem` in its implementation as it would invalidate the cached result created by createSelector as it would change the bag item id parameter.
- Refactored `getBagItems`, `getBagItem`, `getWishlistItems` and `getWishlistItem` selectors to use a common denormalize function.
- Changed `useWishlist` and `useBag` hooks to returned undefined when the wishlist and bag are undefined respectively. This is to avoid generating a new object reference when the data has not changed.
- Changed `renderContent` function to accept only 2 arguments instead
of multiple arguments and made the 2nd argument optional to avoid type
errors when the components registered do not use the same props as the
ones provided by the package.
- Added a `.filter(Boolean)` call in getProductDenormalized selector
for the categories property calculation to avoid receiving an array with
undefined values.

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/bags/selectors.ts
+++ b/packages/redux/src/bags/selectors.ts
@@ -9,14 +9,14 @@ import {
   getItemsIds,
   getResult,
 } from './reducer';
-import { getEntities, getEntityById } from '../entities/selectors';
-import { getProductDenormalized } from '../products/selectors/product';
+import { getEntities } from '../entities/selectors';
 import isEqual from 'lodash/isEqual';
 import omit from 'lodash/omit';
-import type { BagItem, ProductType } from '@farfetch/blackout-client';
+import type { BagItem, Brand, ProductType } from '@farfetch/blackout-client';
 import type {
   BagItemEntity,
   BagItemHydrated,
+  CategoryEntity,
   ProductEntity,
   ProductEntityDenormalized,
 } from '../entities/types';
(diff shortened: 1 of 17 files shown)


packages/redux/src/bags/selectors.ts
```diff
@@ -79,8 +78,38 @@ export const getBagError = (state: StoreState) =>
...
 /**
- * Retrieves a specific bag item by its id, with all properties populated (ie, the
- * product).
+ * Denormalizes a bagItem
+ *
+ * @param bagItemId - The id of bagItem to denormalize
+ * @param bagItems - Bag item entities as obtained with getEntities(state, 'bagItems').
+ * @param products - Product entities as obtained with getEntities(state, 'products').
+ *
+ * @returns Wishlist item denormalized containing the product data if available.
```
anapim, line 87:
```suggestion
 * @returns Bag item denormalized containing the product data if available.
```

anapim (approved):

packages/redux/src/bags/selectors.ts
boliveira: Fixed! Thank you!

andreribeirosilva (approved):

packages/redux/src/bags/selectors.ts
```diff
@@ -96,30 +125,18 @@ export const getBagId = (state: StoreState) => getId(state.bag as BagsState);
...
-// Note: Apparently the type definition of the createSelector function
-//       is not defined correctly in reselect package as it is not inferring
-//       the additional parameter 'bagItemId' provided in each selector, so
-//       we need to type the returned selector ourselves instead of relying on
-//       the inferred type.
 export const getBagItem: (
   state: StoreState,
   bagItemId: BagItem['id'],
-) => BagItemHydrated = createSelector(
+) => BagItemHydrated | undefined = createSelector(
```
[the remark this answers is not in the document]
boliveira: That is not a cast but for the output of createSelector, you are right it is not necessary. However, we need to redefine the type for the `getBagItem` variable as the types inferred for the parameters of createSelector are not easy to understand.

PR #614, opened 2022-12-15
fix(redux|react): fix createSelector types

boliveira:
## Description

- This fixes the typings of all selectors that are based on createSelector as the types generated by createSelector by default are not very understandable for the consumer.
- Changed sharedWishlistItem selectors to avoid sending a new reference even when data has not changed.
- Renamed types that contain `Hydrated` in its name to `Denormalized` to be consistent with other types.
- Fixed Omnitracking integration logging message that did not contain the
Omnitracking prefix.

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/sharedWishlists/selectors.ts
+++ b/packages/redux/src/sharedWishlists/selectors.ts
@@ -1,12 +1,14 @@
 import * as fromSharedWishlistReducer from './reducer';
 import { createSelector } from 'reselect';
-import { getEntityById } from '../entities/selectors';
-import { getProduct } from '../products/selectors/product';
+import { getEntities, getEntityById } from '../entities/selectors';
+import type { Brand, SharedWishlist } from '@farfetch/blackout-client';
 import type {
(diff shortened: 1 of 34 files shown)


packages/redux/src/sharedWishlists/selectors.ts
```diff
@@ -154,22 +157,96 @@ export const getSharedWishlistItem = (
  * ```
  *
  * @param state     - Application state.
- * @param sharedWishlistId - Numeric identifier of the shared wishlist.
```
ruifcnunes, line 157: Should this be removed?

ruifcnunes (approved):

packages/redux/src/sharedWishlists/selectors.ts
boliveira: Yes, my bad. At first I removed it because I was thinking of doing another approach, but then I changed again and forgot to add the jsdoc :(
Fixed, thanks!

PR #646, opened 2023-01-30
feat(core|react): allow analytics to work with any consent categories

ruifcnunes:
## Description
This PR introduces a new feature on the analytics package that allows the projects to define custom consent categories to integrations. The built-in ones will still use its default consent categories.

- Added custom consent feature;
- Adapted the `.shouldLoad()` method on the `Integration` class  to contain all logic regarding the consent and removed this method from the integrations that no longer need it (as all the work will now be done on the static method of the `Integration` class;
- From now on, built-in integrations that need consent can just define a static property on the class called `consentCategories` - this will be used on the `.shouldLoad()` method of the `Integration` class if no custom consent categories are defined via the integration's options.
- Removed some unnecessary static `createInstance` methods from integrations.
- Other minor improvements.

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/analytics/integrations/Integration.js
+++ b/packages/core/src/analytics/integrations/Integration.js
@@ -1,3 +1,4 @@
+import { CONSENT_CATEGORIES_PROPERTY } from '../utils';
 /**
  * Base class for integrations.
  * It ensures the base functionality in order to work with analytics.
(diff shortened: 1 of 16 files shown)


packages/core/src/analytics/integrations/Integration.js
```diff
@@ -19,16 +20,58 @@ class Integration {
...
+   * @param {object} consented - The current consent given by the user.
+   * @param {object} options   - Integration options.
    *
    * @returns {boolean} If the integration is ready to be loaded.
    */
-  // eslint-disable-next-line no-unused-vars
-  static shouldLoad(consent) {
-    return false;
+  static shouldLoad(consented = {}, options) {
+    this.setConsentCategories(options);
```
boliveira, line 62: I am not sure if we should set the consent categories in a static property of the class. Imagine a scenario where I might want 2 instances of the same integration to be loaded with different consent categories. With the static property, the last one whose `shouldLoad` method was run would override the value for all instances.

I know this might be a very peculiar scenario but we should keep in mind that our api allows users to do this:

```js
analytics.addIntegration('int1', MyIntegration, {consentCategories: ['a', 'b']});
analytics.addIntegration('int2', MyIntegration, {consentCategories: ['c', 'd']});
```

If this is valid, then we should not store values in static properties which contain information for each integration instance that was added.

I would suggest removing the `CONSENT_CATEGORIES_PROPERTY` from the `Integration` class and having this instead:

- Add a `defaultConsentCategories` (or `getDefaultConsentCategories` method) to the Integration class which subclasses can override with the default value of the consent categories that will make the integration load.
- In `shouldLoad` do not store the consentCategories from options and use it just as an override of the `defaultConsentCategories`. I believe it is not necessary to store them since it is stored inside the options when `addIntegration` was called.
- Change `setConsentCategories` to `validateConsentCategories` which would only validate if the integration has defined the consent categories for the integration correctly (either via `defaultConsentCategories` or `options.consentCategories`) and it does not need to be inside the Integration class (For instance, a custom integration could override that method, which I guess it shouldn't?)

What do you think? This is a bit theoretical, if you want we can discuss this offline.

packages/core/src/analytics/integrations/__tests__/Integration.test.js
```diff
@@ -29,7 +25,60 @@ describe('Integration', () => {
     expect(integration.track).toBeInstanceOf(Function);
   });
 
-  it('Should have a setConsent method', () => {
-    expect(integration.setConsent).toBeInstanceOf(Function);
+  it('should return an instance of itself or of an extended class', () => {
+    class MyIntegration extends Integration {}
+
+    expect(integration).toBeInstanceOf(Integration);
+    expect(new MyIntegration()).toBeInstanceOf(Integration);
```
boliveira, line 32: I think I do not understand the goal of this test. Like this we are testing the Javascript inheritance mechanism, there isn't a particular behavior that is related to analytics/Integration. Perhaps you meant to test the `createInstance` method?

packages/core/src/analytics/integrations/Integration.js
ruifcnunes: As discussed offline, changed the part where we were overriding the default consent category when we receive a different one from the options - it will only be used for validation and within the `shouldLoad` method to make the decision to load (or not) the integration.
Kept the `consentCategories` naming for "ergonomic" reasons, this way we can use the same key everywhere, keeping the initial goal of not having too much things referring "custom" vs "default" consent.

Let me know if this is okay for you and thanks for the feedback 🙂

packages/core/src/analytics/integrations/__tests__/Integration.test.js
ruifcnunes: yes, it was indeed 😄 fixed, thanks for the heads up!

packages/core/src/analytics/integrations/Integration.js
boliveira: Okay, this is fine. Thanks!

boliveira (approved):

boliveira (approved):

PR #648, opened 2023-01-31
feat(client): add segment at target on searchContents

Edix96:
## Description

Add `target.segment ` on searchContents function

## Checklist

## Disclaimer

### DIFF
--- a/packages/client/src/contents/types/contents.types.ts
+++ b/packages/client/src/contents/types/contents.types.ts
@@ -6,6 +6,7 @@ export type Targets = {
   country?: string;
   benefits?: string;
   channel?: string;
+  segments?: string;
 } & Record<string, string | undefined>;
 
 export type QuerySearchContents = {
@@ -17,7 +18,7 @@ export type QuerySearchContents = {
   contentTypeCode: string;
   // List of codes that representing the content code (about-us|today-news|header|productId...).
   codes?: string | string[];
-  // The targets and respective values that a content type is configured (contentzone:ROW | country:GB | language:en-GB | benefits:test).
+  // The targets and respective values that a content type is configured (contentzone:ROW | country:GB | language:en-GB | benefits:test | segments: 'private-sale,private-sale-guest-user').  // The segments list should contain less than 10 strings. They should separated by a comma (,).
   target?: Targets;
   // Sort content by (publicationDate:desc | publicationDate:asc | metadataCustom.eventDate:desc | metadataCustom.X:asc).
   sort?: string;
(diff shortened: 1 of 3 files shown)


joaoprcosta (approved):

packages/client/src/contents/types/contents.types.ts
```diff
@@ -6,7 +6,8 @@ export type Targets = {
   country?: string;
   benefits?: string;
   channel?: string;
-} & Record<string, string | undefined>;
+  segments?: string[];
```
boliveira, line 9: If segments can have at most 10 elements, what do you think of defining this type as a tuple with conditional indices, like this:

```js
segments?: [string, string?, string?, string?, string?, string?, string?, string?, string?, string?]
```

boliveira: (I am assuming that the first position will always exist if the `segments` property exists in the payload but feel free to change it to optional as well if that is not the case)

Edix96: I understand your suggest, but this type wouldn’t be too long?

boliveira: That is not a problem if it is correct and helps users catch errors. Is it true that it will only have at most 10 segments? If it is, then do not have a problem with the type being too long.

boliveira: You can also define a new type alias for that and use that instead like this:

```js
export type TargetSegments = [string, string?, string?, string?, string?, string?, string?, string?, string?, string?];
...
segments?: TargetSegments;
```

Edix96: In conversation with Ricardo, they decided that the segments should be passed as a string where the segments are seperated by commas

packages/client/src/contents/types/contents.types.ts
```diff
@@ -17,7 +18,8 @@ export type QuerySearchContents = {
   contentTypeCode: string;
   // List of codes that representing the content code (about-us|today-news|header|productId...).
   codes?: string | string[];
-  // The targets and respective values that a content type is configured (contentzone:ROW | country:GB | language:en-GB | benefits:test).
+  // The targets and respective values that a content type is configured (contentzone:ROW | country:GB | language:en-GB | benefits:test | segments: ['private-sale']).
```
boliveira, line 21: I think this comment is stale. Can you update it to something like this?

```suggestion
  // The targets and respective values that a content type is configured (contentzone:ROW | country:GB | language:en-GB | benefits:test | segments: 'private-sale, other-segment').
```

Edix96: Thanks!

vinicaval (approved):

PR #649, opened 2023-01-31
feat(react): Add forter and riskified integrations

danielbento92:
## Description

- Add Forter and Riskified integrations;
- Remove vitorino;
- Refactor of current tests and add new ones.

throws an error and added a unit test for it.
- Changed `skipLibCheck` to false to allow type checking of `.d.ts` files
in the project.
- Added `@types/jquery` package to allow typing calls to jquery.
- Removed packages/redux/tsconfig.ci.tests.json as it was not needed
anymore.
- Added `dist` folder exclusion to `tsconfig.ci.tests.json` since
now with `skipLibCheck` set to false, it would catch the files from
`dist` folders and would give errors if they were available from a
 previous compilation.

BREAKING CHANGE: Vitorino was removed from the list of available
integrations and now there are separate Riskified and Forter integrations
that will need to be added separately to have the same behaviour
provided by Vitorino.

### Dependencies

None.

## Checklist

## Disclaimer


packages/analytics/src/integrations/Integration.ts
```diff
@@ -59,7 +59,7 @@ class Integration<T extends IntegrationOptions> {
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   track(this: this, data: EventData<TrackTypesValues>): void {
-    throw new Error('Method not implemented');
+    // Do nothing
```
ruifcnunes, line 62: I think we should keep this throw here. If this is causing issues on integrations that do not need to track anything, just override the method there to specifically do nothing and with that, the error will not appear.

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.baseUrl = `https://beacon.riskified.com?shop=${this.siteId}&sid=`;
+
+    this.setupScript(loadData);
+  }
+
+  /* static createInstance<Options extends IntegrationOptions>(
+    this: IntegrationFactory<Options>,
+    options: Options,
+*/
+  static override createInstance<Options extends RiskifiedIntegrationOptions>(
```
ruifcnunes, line 42: This method is not needed as the `Integration` class already does the same.

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * When the script loads onto the DOM, it will set the integration as ready.
+   */
+  onLoad = () => {
+    this.isReady = true;
+  };
+
+  getOptions() {
```
ruifcnunes, line 128: is there a use-case for these getters here?

packages/react/src/analytics/integrations/Vitorino/providers/__tests__/Forter.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    expect(
+      Forter.createInstance(
+        { siteId: mockSiteId },
+        loadIntegrationData,
+        strippedDownAnalytics,
+      ),
+    ).toBeInstanceOf(Forter);
+  });
+
+  it('should return default `siteId` if empty value are passed on options', () => {
```
ruifcnunes, line 79:
```suggestion
  it('should return default `siteId` if an empty value is passed on options', () => {
```
(also valid for test descriptions below)

packages/analytics/src/integrations/index.ts
```diff
@@ -1,2 +1,6 @@
+import { formatPageEvent, formatTrackEvent } from './Omnitracking';
+
+export const OmnitrackingHelpers = { formatPageEvent, formatTrackEvent };
```
boliveira, line 3: Why is this being exported here?

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -10,59 +10,54 @@
...
-   */
-  static override shouldLoad() {
-    return true;
-  }
+class Vitorino extends integrations.Integration<VitorinoIntegrationOptions> {
+  private forter?: Forter;
+  private riskified?: Riskified;
+  private baseOptions: VitorinoIntegrationBaseOptions;
+
+  // /**
```
boliveira, line 45: Can these comments be removed?

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -72,204 +67,87 @@ export default class Vitorino extends integrations.Integration<VitorinoIntegrati
...
   ) {
+    const baseOptions: VitorinoIntegrationBaseOptions = omit(options, [
+      'riskified',
+      'forter',
+    ]);
     super(options, loadData, analytics);
 
-    this.isScriptLoaded = false;
+    this.baseOptions = baseOptions;
+    this.options = options;
```
boliveira, line 81: This is not necessary since the super will take care of that.

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -72,204 +67,87 @@ export default class Vitorino extends integrations.Integration<VitorinoIntegrati
...
-   *
-   * @param options - Integration options.
-   */
-  buildMapper(options: VitorinoIntegrationOptions) {
-    const customMapperFn = options.eventsMapper;
-
-    if (customMapperFn && typeof customMapperFn !== 'function') {
-      utils.logger.error(INVALID_CUSTOM_MAPPER_ERROR_MESSAGE);
+    // load default if not set
+    options.activeIntegrations ??= [
```
boliveira, line 86: Careful, you are setting a property in an object that is being shared with the code that called `addIntegration`. What do you think of setting it as a prop of the class?

boliveira: (or local variable)

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -10,59 +10,54 @@
...
-   * This integration is required, so it must be loaded independently of user
-   * consent.
-   *
-   * @returns If the integration should load or not.
-   */
-  static override shouldLoad() {
-    return true;
-  }
+class Vitorino extends integrations.Integration<VitorinoIntegrationOptions> {
+  private forter?: Forter;
```
boliveira, line 41: I think `forterInstance` would be a better name. What do you think? (Same for `riskified` prop)

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -72,204 +67,87 @@ export default class Vitorino extends integrations.Integration<VitorinoIntegrati
...
-  getEnvironmentFromOptions(options: VitorinoIntegrationOptions) {
-    const { environment } = options;
-    const safeEnvironment =
-      environment || window?.__BUILD_CONTEXT__?.env?.NODE_ENV;
-
-    return safeEnvironment === ENVIRONMENT_TYPES.prod
-      ? window.Vitorino.Environment.PRODUCTION
-      : window.Vitorino.Environment.SANDBOX;
+  override async onSetUser(data: SetUserEventData): Promise<void> {
+    await this.forter?.onSetUser(data);
```
boliveira, line 149: Can this be done in parallel instead of calling forter's onSetUser first and only then calling riskified's onSetUser? Also, do they need to be awaited? If so, and they can be done in parallel, use `Promise.all()` instead.

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -282,41 +160,25 @@ export default class Vitorino extends integrations.Integration<VitorinoIntegrati
    * @returns Promise that will resolve when the method finishes.
    */
   override async track(data: EventData<TrackTypesValues>) {
-    await this.initializePromise;
-    await this.userIdPromise;
-
-    try {
-      const PageTypes = this.getPageTypeFromEvent(data.event);
-      let safePageTypes;
+    await this.forter?.track(data);
```
boliveira, line 163: Same as https://github.com/Farfetch/blackout/pull/649/files#r1092307689

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -282,41 +160,25 @@ export default class Vitorino extends integrations.Integration<VitorinoIntegrati
...
+    return this.baseOptions;
+  }
 
-    if (this.options.debugger) {
-      utils.logger.info('Vitorino track: ', {
-        config: this.config,
-        vitorinoPageType: page,
-      });
-    }
+  getOptions() {
```
boliveira, line 179: Why expose the same options that were passed from `addIntegration`?

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -282,41 +160,25 @@ export default class Vitorino extends integrations.Integration<VitorinoIntegrati
...
-   * Calls the callback function given from the config call to register a page view
-   * change.
-   *
-   * @param page - The `page` parameter of the event.
-   */
-  callVitorino(page?: string) {
-    if (this.isVitorinoConfigured && this.vitorinoTrackCallback) {
-      this.vitorinoTrackCallback(page);
-    }
+  getBaseOptions() {
```
boliveira, line 175: Is this necessary to be exposed as well?

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,149 @@
...
+  private isReady: boolean;
+  private siteId: string;
+  private baseUrl: string;
+
+  /**
+   * Creates an instance of Riskified and sets the script on the page.
+   *
+   * @param options - Custom options for the integration.
+   * @param loadData - Analytics's load event data.
+   */
```
boliveira, line 23: Missing analytics parameter

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,149 @@
...
+  constructor(
+    options: RiskifiedIntegrationOptions,
+    loadData: LoadIntegrationEventData,
+    analytics: StrippedDownAnalytics,
+  ) {
+    super(options, loadData, analytics);
+
+    this.isReady = false;
+    this.siteId = options.siteId || 'farfetch.com';
+    this.baseUrl = `https://beacon.riskified.com?shop=${this.siteId}&sid=`;
```
boliveira, line 33: I am not sure if this should be set on the source code due to security concerns. I would have this option being passed by the consumer in `analytics.addIntegration` and have documentation that explains that this is the value that should be set.

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,149 @@
...
+   */
+  constructor(
+    options: RiskifiedIntegrationOptions,
+    loadData: LoadIntegrationEventData,
+    analytics: StrippedDownAnalytics,
+  ) {
+    super(options, loadData, analytics);
+
+    this.isReady = false;
+    this.siteId = options.siteId || 'farfetch.com';
```
boliveira, line 32: See https://github.com/Farfetch/blackout/pull/649/files#r1092316527

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.baseUrl = `https://beacon.riskified.com?shop=${this.siteId}&sid=`;
+
+    this.setupScript(loadData);
+  }
+
+  /* static createInstance<Options extends IntegrationOptions>(
```
boliveira, line 38: Can this be removed?

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,149 @@
...
+   * @param data - The event object.
+   *
+   * @returns The Riskified instance to allow chaining calls.
+   */
+  override track(data: EventData<TrackTypesValues>) {
+    switch (data.type) {
+      case TrackTypes.PAGE:
+        return this.trackPage(get(data, 'context.web.window.location.href'));
+
+      case TrackTypes.TRACK:
```
boliveira, line 96: This case can be removed right?

packages/react/src/analytics/integrations/Vitorino/types/types.ts
```diff
@@ -2,6 +2,8 @@ import type {
   EventContext,
   IntegrationOptions,
 } from '@farfetch/blackout-analytics';
+import type { ForterOptions } from '../providers/Forter/types';
+import type { RiskifiedOptions } from '../providers/Riskified/types';
 
 export type EventsMapper = Record<
```
boliveira, line 8: There are some types here like `Fields`, `Network` and `Environment` that are not being exported, but should as they are being used in other public exported types.

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/types/types.ts
```diff
@@ -0,0 +1,8 @@
+import type { VitorinoIntegrationBaseOptions } from '../../../types';
+
+export interface RiskifiedOptions {
+  siteId?: string;
```
boliveira, line 4: I noticed both `RiskifiedOptions` and `ForterOptions` have the `siteId` property. If they are supposed to have the same value, shouldn't this option be in `VitorinoIntegrationBaseOptions`?

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * Tracks the page view event. Calls the global Riskified variable RISKX.go with the web location.
+   *
+   * @param location - The location data.
+   *
+   * @returns The Riskified instance to allow chaining calls.
+   */
+  trackPage(location: string): this {
```
boliveira, line 111: Track page is returning the instance but afaics it is not being used anywhere. This method is being called by the track method which is called by analytics which does not do anything with the return value.

packages/react/src/analytics/integrations/Vitorino/providers/__tests__/Forter.test.ts
```diff
@@ -0,0 +1,177 @@
...
+// @ts-expect-error
+const strippedDownAnalytics: StrippedDownAnalytics = {
+  createEvent: (type: TrackTypes) =>
+    Promise.resolve({ ...loadIntegrationData, type }),
+};
+jest.mock('@farfetch/blackout-client', () => {
+  const defaultClient = jest.requireActual('@farfetch/blackout-client');
+
+  return {
+    UserGender: defaultClient.UserGender,
```
boliveira, line 41: This override seems unnecessary.

packages/react/src/analytics/integrations/Vitorino/constants.ts
```diff
@@ -1,45 +1,4 @@
...
-  [PageTypes.LOGIN_REGISTER]: [
-    window.Vitorino.PageTypes.REGISTER,
-    window.Vitorino.PageTypes.LOGIN,
-  ],
-  [EventTypes.PLACE_ORDER_STARTED]: window.Vitorino.PageTypes.CHECKOUT,
-  [EventTypes.SIGNUP_FORM_VIEWED]: window.Vitorino.PageTypes.REGISTER,
-  [EventTypes.CHECKOUT_STEP_VIEWED]: window.Vitorino.PageTypes.SHIPPING,
-});
+export enum VitorinoProviders {
+  RISKIFIED = 'riskified',
```
boliveira, line 2: Prefer camel case here, like `Riskified`.
Also this enum should be in `types` and not in `constants` file.

packages/react/src/analytics/integrations/Vitorino/providers/Forter/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const ForterTokenReadyEvent = 'ftr:tokenReady';
+export const ForterTokenTid = 1364;
+export const ForterTokenLoadedAnalyticsEvent = 'forterTokenLoaded';
+
+// site ID for forter sandbox - https://portal.forter.com/api/general/introduction
+export const ForterSandboxSiteId = 'b2f674e65110';
+// site ID for forter production - https://portal.forter.com/api/general/introduction
+export const ForterProductionSiteId = '0c0275c12bca';
```
boliveira, line 8: Same as with the Riskified comment, this should not be exposed in source code. Have the user pass it and document which values he should use.

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/types/types.ts
boliveira: Never mind, just realized that they are supposed to be different.

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
```diff
@@ -10,59 +10,54 @@
  * ```
  */
```
boliveira, line 11: The tsdoc for this class should be updated as well.

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: This is done already on Vitorino script. The objective of this initiative it's to deprecate the usage of Vitorino script, maintaining behavior but from blackout side. Besides that, yes, these keys could be removed from integration and set only by addIntegration, but it needs to be discussed because it doesn't be aligned with task requirements.

packages/analytics/src/integrations/Integration.ts
danielbento92: I made this change because as we talked offline, Vitorino would trigger track or onSetUser events, for each active integration, which only one integration using only onSetUser, and the other using only track method.
With this throw, the integration that won't override this track method (the one that won't use the track method), will trigger this error by triggering the track from the main class.

packages/analytics/src/integrations/index.ts
danielbento92: Forter uses FormatTrackEvent when it trigger, internally, an omnitracking event.

boliveira: Right, when I searched for it in VSCode somehow it was not finding it but I can see it now.
In this case, I would suggest exposing these methods as static inside the `Omnitracking` integration class, then consumers who needed it would use them like this:

```js
const formattedEvent = Omnitracking.formatPageEvent(event)
```

And then you can get rid of these exports as they "pollute" the integrations export with other things that are not integrations.

packages/analytics/src/integrations/Integration.ts
danielbento92: reverted to throw, plus added track on missing Forter class (without any functionality associated)

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: Removed

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: Removed and refactored the tests

packages/react/src/analytics/integrations/Vitorino/providers/__tests__/Forter.test.ts
danielbento92: This test was removed.

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: Was created a new variable for activeIntegrations instead of using previous from options directly.

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: I agree and applied.

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: Fixed.

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: Removed

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: Removed

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: Removed

packages/react/src/analytics/integrations/Vitorino/types/types.ts
danielbento92: Added export for this types

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: Removed the return in this method.

packages/react/src/analytics/integrations/Vitorino/providers/__tests__/Forter.test.ts
danielbento92: Applied.

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/Vitorino/Vitorino.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/Vitorino/constants.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/Guardian/Guardian.ts
```diff
@@ -0,0 +1,146 @@
...
+      GuardianProviders.Forter,
+      GuardianProviders.Riskified,
+    ];
+
+    // Check if activeIntegrations is an array
+    if (!Array.isArray(this.activeIntegrations)) {
+      utils.logger.error(
+        '[Analytics] Guardian - The value `activeIntegrations` from Guardian integration options must be an array.',
+      );
+      return;
```
boliveira, line 78: This should throw an error since the instance will not be in the correct state. This will make any call to `track`, `onSetUser` to fail later when the integration should be removed from the active integrations in analytics.

packages/react/src/analytics/integrations/Guardian/Guardian.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    if (this.activeIntegrations.includes(GuardianProviders.Forter)) {
+      this.forterInstance = Forter.createInstance(
+        getSafeOptionsForIntegration<ForterIntegrationOptions>(
+          baseOptions,
+          options,
+          'forter',
```
boliveira, line 86: This could use the `GuardianProviders.Forter` as well (the same for `riskified` below).

packages/react/src/analytics/integrations/Guardian/Guardian.ts
boliveira: Another question: Since the user has also to pass the `riskified` and `forter` options on the integration options for them to initialize is it necessary to have the `activeIntegrations` option as well? Can't we derive the active integrations from the `riskified` and `forter` options?

packages/react/src/analytics/integrations/Guardian/Guardian.ts
```diff
@@ -0,0 +1,146 @@
+/**
+ * Guardian Integration. Loads the Guardian script and calls its methods.
+ *
+ * @example <caption>Adding Guardian integration to analytics</caption>
```
boliveira, line 4: This description is not correct since this integration does not load a script as Vitorino once did.

packages/react/src/analytics/integrations/Guardian/Guardian.ts
```diff
@@ -0,0 +1,146 @@
+/**
+ * Guardian Integration. Loads the Guardian script and calls its methods.
```
boliveira, line 2: This description is not correct since this integration does not load a script as Vitorino once did.

packages/react/src/analytics/integrations/Guardian/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * @returns A value indicating if the integration should load.
+   */
+  static override shouldLoad(): boolean {
+    return true;
+  }
+
+  /**
+   * Creates the script based on the correlation `id` and appends it on the page.
+   * Once the script is on the page, it will automatically call `RISKX.go()`.
```
boliveira, line 49: I think this description is not correct. Once the script is on the page, it will not call `RISKX.go` but will only set the `isReady` property to true. Can you check, please?

packages/react/src/analytics/integrations/Guardian/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Tracks the page view event. Calls the global Riskified variable RISKX.go with the web location.
+   *
+   * @param location - The location data.
+   *
+   */
+  trackPage(location: string) {
+    if (!this.isReady) {
+      return;
```
boliveira, line 92: Related with https://github.com/Farfetch/blackout/pull/649/files#r1097218215, since it is returning here when it is not ready, when it is ready, it should call `RISKX.go` (which would make the description for the `setupScript` correct) since the event will be lost. I am not sure if instead of just returning, the current event data should be added to a queue, which would then be emptied when the script was loaded. Please, take a look at it since the way it is now, some page tracks can be lost.

packages/react/src/analytics/integrations/Vitorino/types/types.ts
boliveira: Sorry to comment again on this, but could you add a more specific name for these types? `Fields`, `Network` and `Environment` are too generic and these types are specific to Guardian. Also, there is an `EnvironmentTypes` type as well which is being used to compare to the value of the `environment` option in Guardian which is typed as `Environment`. My question is can't it reuse the `EnvironmentTypes` and drop `Environment` altogether?

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      // @ts-expect-error
+      const guardianInstance = Guardian.createInstance() as Guardian;
+
+      expect(guardianInstance.activeIntegrations).toEqual([
+        GuardianProviders.Forter,
+        GuardianProviders.Riskified,
+      ]);
+
+      expect(guardianInstance.riskifiedInstance).toBeInstanceOf(
+        integrations.Integration,
```
boliveira, line 47: Shouldn't this test if it is an instance of Riskified? Same for the test of `forterInstance` below.

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        activeIntegrations: [GuardianProviders.Forter],
+      }) as Guardian;
+
+      expect(guardianInstance.riskifiedInstance).toBeUndefined();
+      expect(guardianInstance.forterInstance).toBeInstanceOf(
+        integrations.Integration,
+      );
+    });
+
+    it('Should trigger an error', () => {
```
boliveira, line 66: Be more explicit as to why this should trigger an error, please.

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      Guardian.createInstance({
+        activeIntegrations: {},
+      }) as Guardian;
+
+      expect(spyError).toHaveBeenCalledWith(
+        '[Analytics] Guardian - The value `activeIntegrations` from Guardian integration options must be an array.',
+      );
+    });
+
+    it('Should trigger warn about invalid integration settings', () => {
```
boliveira, line 77:
```suggestion
    it('Should trigger warnings about invalid integration settings', () => {
```

packages/react/src/analytics/integrations/Guardian/Guardian.ts
```diff
@@ -0,0 +1,146 @@
...
+    // load default if not set
+    this.activeIntegrations = baseOptions.activeIntegrations || [
+      GuardianProviders.Forter,
+      GuardianProviders.Riskified,
+    ];
+
+    // Check if activeIntegrations is an array
+    if (!Array.isArray(this.activeIntegrations)) {
+      utils.logger.error(
+        '[Analytics] Guardian - The value `activeIntegrations` from Guardian integration options must be an array.',
```
boliveira, line 76: Also be consistent with the error messages. Other integrations do not prefix `[Analytics]` to their error messages and just prefix the integration by itself like `[Guardian]`. Take a look at other integrations error messages and use the same format, please.

packages/react/src/analytics/integrations/Guardian/guardian-helper.ts
```diff
@@ -0,0 +1,60 @@
...
+  integration: keyof GuardianProviderOptions,
+): T {
+  const subOption = fullOptions[integration];
+  if (subOption && isObject(subOption) && !Array.isArray(subOption)) {
+    return { ...baseOptions, ...subOption } as T;
+  }
+
+  if (subOption) {
+    utils.logger.warn(
+      `[Analytics] Guardian - The value of ${integration} from Guardian integration options must be an object with ${integration} integration options.`,
```
boliveira, line 55:
```suggestion
      `[Guardian] - The value of '${integration}' from Guardian integration options must be an object with '${integration}' integration options.`,
```

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  const defaultKey = 'farfetch.com';
+  const baseUrl = (key?: string) =>
+    `https://beacon.riskified.com?shop=${key || defaultKey}&sid=`;
+  const localId = '3c11a6f9-6d5e-45c4-b1b2-d26736964605';
+
+  const loadData = {
+    user: {
+      localId,
+    },
+  };
```
boliveira, line 14: Add an empty line here please

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
```diff
@@ -0,0 +1,125 @@
+import { EnvironmentTypes, integrations } from '@farfetch/blackout-analytics';
+import { Riskified } from '..';
+
+describe('Riskified', () => {
+  const defaultKey = 'farfetch.com';
+  const baseUrl = (key?: string) =>
+    `https://beacon.riskified.com?shop=${key || defaultKey}&sid=`;
+  const localId = '3c11a6f9-6d5e-45c4-b1b2-d26736964605';
+
+  const loadData = {
```
boliveira, line 10: Instead of creating a new loadData mock, there is a fixture in analytics for this exact reason in `loadIntegrationData.fixtures.ts` that you can reuse here.

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      };
+
+      instance.isReady = true;
+      // @ts-expect-error
+      instance.track(data);
+
+      expect(spy).toBeCalledWith(data.context.web.window.location.href);
+    });
+
+    it('Should not call `RISKX.go` if the integration isnt ready', () => {
```
boliveira, line 87:
```suggestion
    it('Should not call `RISKX.go` if the integration is not ready', () => {
```

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        go: spy,
+      };
+
+      instance.isReady = false;
+      // @ts-expect-error
+      instance.track(data);
+
+      expect(spy).not.toBeCalled();
+    });
+  });
```
boliveira, line 111: Add an empty line here, please.

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  describe('Riskified Integration without defaults', () => {
+    it('Should use not default siteId', () => {
```
boliveira, line 113:
```suggestion
    it('Should not use the default siteId if one is passed in options', () => {
```

packages/react/src/analytics/integrations/Guardian/Guardian.ts
danielbento92: Riskified and Forter do not currently require any parameters as they are all optional. The aim is to allow users to either use the default settings or customize them if needed.

By filling out the "Riskified" or "Forter" object in GuardianOptions, it may require additional code for tenants who only want to set up the default options.

This way, it becomes easy to understand that there is one property to customize the enabled integrations and any custom properties for each integration can be specified in their respective "sub-option object".

packages/react/src/analytics/integrations/Guardian/Guardian.ts
danielbento92: I believe the purpose of this enum is not exactly the same as what is requested in the method, however, it could still be utilized as it aligns with the necessary parameters.
The idea behind the enum is to create a list of all possible providers, while the method aims to retrieve specific options for a given integration by passing the intended "subOptions" so that it can concatenate the mainOptions.

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
danielbento92: I was simply verifying if all instances are automatically loaded. I believe it would be enought to ensure that instances designated as `riskified` or `forter` are indeed instances of `Integration` which is the class where each one extend.

So I don't see here too much importance to validate the exact class type in Guardian, but I will take care to include an additional test for this purpose.

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
danielbento92: fixed

packages/react/src/analytics/integrations/Guardian/providers/Riskified/Riskified.ts
```diff
@@ -0,0 +1,106 @@
...
+   * @param loadData - Analytics's load event data.
+   *
+   * @returns The Riskified instance to allow chaining calls.
+   */
+  setupScript(loadData: LoadIntegrationEventData): this {
+    const correlationId = get(loadData, 'user.localId');
+    const script = document.createElement('script');
+    const url = `${this.baseUrl}${correlationId}`;
+
+    document.body.appendChild(script);
```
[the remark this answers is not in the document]
danielbento92: fixed

packages/react/src/analytics/integrations/Guardian/providers/Riskified/Riskified.ts
danielbento92: fixed

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
danielbento92: Now without defaults to Shop property, this case will be deleted

packages/react/src/analytics/integrations/Guardian/Guardian.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
boliveira: Just for information: The test should indeed check for the correct instance type, even though typescript could catch these errors but it might not if for example, the types of both `Forter` and `Riskified` classes are compatible (for example if `Riskified` was a subclass of `Forter` or they had the same property names and types). If that were the case, the test would not detect a bug if the code set the Forter instance to a Riskified instance and vice-versa.

danielbento92: As I tried to explain, in this specific test I only are validating if instances are set or not depending of activeIntegration property.
Then I added an additional case, to really test if forter instance are really a forter, and same for riskified.

So in my opinion, this could be done in two different tests for different proposes (one for validating if instance aren't undefined, and another to validate the type of each instance), instead of mixing in a single test.

boliveira: Imagine that you have a bug in your code that if `riskified` option is passed, you are creating a `forter` instance instead. The test the way it is now, would not catch it. I do not see the need of having a separate test for the type of instances when you can have this test do it and would catch more errors. The test you added for that is only testing when no options are being sent (which is not valid now since both providers will need to receive options in order to be instantiated).

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: Update: It was removed the defaults, as we discuss offline.

packages/react/src/analytics/integrations/Vitorino/providers/Riskified/Riskified.ts
danielbento92: Update: It was removed the defaults, as we discuss offline.

packages/react/src/analytics/integrations/Vitorino/types/types.ts
danielbento92: Changed to use EnvironmentTypes instead of previous Environment

packages/react/src/analytics/integrations/Vitorino/providers/Forter/constants.ts
danielbento92: Update: It was removed the defaults, as we discuss offline.

packages/react/src/analytics/integrations/Guardian/Guardian.ts
danielbento92: Updated

packages/react/src/analytics/integrations/Guardian/providers/Riskified/Riskified.ts
danielbento92: I've added a promise to track the page load to ensure that events are captured properly. This way, we can guarantee that the events will only run after the script has been fully loaded, and none event will be lost.

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
danielbento92: Updated

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
danielbento92: fixed

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
danielbento92: fixed

packages/react/src/analytics/integrations/Guardian/guardian-helper.ts
danielbento92: fixed

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
danielbento92: fixed

packages/react/src/analytics/integrations/Guardian/providers/__tests__/Riskified.test.ts
danielbento92: fixed

packages/react/src/analytics/integrations/Vitorino/types/types.ts
boliveira: `Fields` and `Network` are still using the same names and they are too generic to be exported as is. From what I can see they are used only inside the `Config` type (which is also too generic). It should be something like this:

`Config` -> `GuardianConfig` (also double check if Config is being used because as far as I can see it is not being used anywhere.
`Fields` -> `GuardianSensitiveFields`
`Network` -> `GuardianNetworkSettings`

packages/react/src/analytics/integrations/Guardian/Guardian.ts
boliveira: Ok, see my previous commentary regarding the `GuardianProviders` enum, which I believe can be removed altogether.

packages/react/src/analytics/integrations/Guardian/providers/Riskified/Riskified.ts
boliveira: Thanks!

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    });
+  });
+
+  describe('Guardian instance', () => {
+    const spyForter = jest.spyOn(Forter, 'createInstance');
+    const spyRiskified = jest.spyOn(Riskified, 'createInstance');
+
+    beforeEach(jest.clearAllMocks);
+
+    it('should not load all instances by default', () => {
```
boliveira, line 37: I think the test description is contrary to what the test does. Can you check, please?

packages/react/src/analytics/integrations/Guardian/__tests__/Guardian.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    it('should only load Forter instance', () => {
+      // @ts-expect-error
+      const guardianInstance = Guardian.createInstance({
+        activeIntegrations: [GuardianProviders.Forter],
+      }) as Guardian;
+
+      expect(guardianInstance.riskifiedInstance).toBeUndefined();
+      expect(guardianInstance.forterInstance).toBeInstanceOf(
+        integrations.Integration,
```
boliveira, line 58: As discussed in another comment, please use the `Forter` class in the instanceof check please.

boliveira (approved):

ruifcnunes (approved):

packages/react/src/analytics/integrations/Forter/__tests__/Forter.test.ts
```diff
@@ -0,0 +1,209 @@
...
+const mockPostTrackingsPromise = new Promise(
+  resolve => (mockPostTrackingsPromiseResolve = resolve),
+);
+const mockSiteId = 'c3f674e6511e';
+const mockOrigin = 'Whitelabel Portal';
+const mockForterToken =
+  '44807307964b4931a89bf364e580bdc5_1594724271711__UDF43_9ck';
+const mockForterTokenLoadedPromise = Promise.resolve(mockForterToken);
+
+// @ts-expect-error
```
boliveira, line 27: Why is this `@ts-expect-error` necessary? It is a good practice to describe why you needed to have it so we can understand why you needed to do so. If I remove this the only error I got was a missing import of the `StrippedDownAnalytics` type that when adding an import to it, cleared the error.

packages/react/src/analytics/integrations/Forter/__tests__/Forter.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        siteId: mockSiteId,
+        origin: mockOrigin,
+      },
+      loadIntegrationData,
+      { ...strippedDownAnalytics, createEvent: createEventMock },
+    );
+    // @ts-expect-error accessing private property just for testing
+    const spy = jest.spyOn(forterInstance, 'resolveOnSetUserPromise');
+    // @ts-expect-error test data
+    const data: SetUserEventData = {};
```
boliveira, line 114: There is an `onSetUserEventData` mock you can import and use here. Using an event that is not correctly mocked can skew the tests. Change it to be like this, please and you can then remove the `@ts-expect-error` from the line before:

```suggestion
    const data: SetUserEventData = onSetUserEventData;
```

packages/react/src/analytics/integrations/Forter/__tests__/Forter.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        customerId: `${eventData.user.id}`,
+        event: 'PageAction',
+        parameters: expect.objectContaining({
+          clientTimestamp: expect.any(String),
+          tid: ForterTokenTid,
+          uuid: expect.any(String),
+          val: JSON.stringify({
+            forterTokenCookie: mockForterToken,
+            origin: mockOrigin,
+            // @ts-expect-error
```
boliveira, line 134: Same as https://github.com/Farfetch/blackout/pull/649/files#r1105523379, add a description for why you had to add this, please.

packages/react/src/analytics/integrations/Forter/__tests__/Forter.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      event: string,
+    ) {
+      return Promise.resolve({
+        ...eventData,
+        event,
+        type,
+      });
+    };
+
+    jest.clearAllMocks();
```
boliveira, line 160: This is unnecessary since there is a `beforeEach` block that will call `jest.clearAllMocks` already.

packages/react/src/analytics/integrations/Forter/__tests__/Forter.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      {
+        siteId: mockSiteId,
+        origin: mockOrigin,
+        omnitrackingHttpClient: httpClientMock,
+      },
+      loadIntegrationData,
+      { ...strippedDownAnalytics, createEvent: createEventMock },
+    );
+    // @ts-expect-error test data
+    const data: SetUserEventData = {};
```
boliveira, line 181: Same as pointed out in another comment, use `onSetUserEventData` export.

packages/react/src/analytics/integrations/Forter/Forter.ts
```diff
@@ -0,0 +1,192 @@
...
+   */
+  async sendForterTokenLoadedEventForOmnitracking(forterToken: string) {
+    await this.onSetUserPromise;
+
+    const omnitrackingMessage = await this.getForterTokenLoadedEventPayload(
+      forterToken,
+    );
+
+    if (this.options.omnitrackingHttpClient) {
+      await this.options.omnitrackingHttpClient(omnitrackingMessage);
```
boliveira, line 112: I think we should handle failures for this call and the call below from `postTracking` since some browsers might not show a warning if this happens (and in node 16 it will crash the process altogether)

packages/react/src/analytics/integrations/Forter/Forter.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * Tracks events (page views or page actions).
+   *
+   * @param _data - The event object.
+   *
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  override track(_data: EventData<TrackTypesValues>) {
```
boliveira, line 187:
```suggestion
  override track() {
```

With this, you can safely remove the `eslint-disable-next-line` rule

packages/analytics/src/types/EnvironmentTypes.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Analytics tracking environment types. These types are meant to be used whenever a call
+ * is made to an integration, to send the proper type of event on the payload.
+ */
+enum EnvironmentTypes {
```
boliveira, line 5: Does this enum still make sense since it is not being used anymore?

packages/react/src/analytics/integrations/Forter/__tests__/forterTokenLoadedEventDetector.test.ts
```diff
@@ -0,0 +1,60 @@
+// @ts-nocheck
```
boliveira, line 1: Why was this necessary?

packages/react/src/analytics/integrations/Forter/types/globals.d.ts
```diff
@@ -0,0 +1,3 @@
+interface Window {
+  jQuery: JQueryStatic;
```
boliveira, line 2: Is this doing anything?

packages/react/src/analytics/integrations/Riskified/Riskified.ts
```diff
@@ -0,0 +1,113 @@
...
+  TrackTypesValues,
+  utils,
+} from '@farfetch/blackout-analytics';
+import get from 'lodash/get';
+import type { RiskifiedIntegrationOptions } from './types';
+
+class Riskified extends integrations.Integration<RiskifiedIntegrationOptions> {
+  baseUrl!: string;
+  isReadyPromise: Promise<unknown>;
+  resolveIsReadyPromise?: (value?: unknown) => void;
```
boliveira, line 16: Since `resolveIsReadyPromise` will be set in the constructor, you can change that to this:

```suggestion
  resolveIsReadyPromise!: (value?: unknown) => void;
```

And remove line 32.

packages/react/src/analytics/integrations/Riskified/Riskified.ts
```diff
@@ -0,0 +1,113 @@
...
+  ) {
+    super(options, loadData, analytics);
+
+    this.resolveIsReadyPromise = undefined;
+    this.isReadyPromise = new Promise(resolve => {
+      this.resolveIsReadyPromise = resolve;
+    });
+
+    if (!options.shop) {
+      utils.logger.warn(
```
boliveira, line 38: I am not sure if I have talked about this before or not (my apologies if I have not) but we should throw in this case since without the `shop` property this instance will not have set up itself correctly, so it should not take space in the `activeIntegrations` array in analytics.

packages/react/src/analytics/integrations/Forter/types/globals.d.ts
danielbento92: Yes, it was used on setupForterTokenLoadedEventListner to type window['jquery'] as JQueryStatic.

boliveira: Fixed by adding `@types/jquery` and changing `skipLibCheck` to false to allow type checking of `.d.ts` files

ruifcnunes (approved):

PR #652, opened 2023-02-02
fix(redux): fix product attributes server render

boliveira:
## Description

This changes the serverInitialState handling of the
`productAttributes` property from page model to discard it if the value
is an empty array since the server will return an empty array by default
even when no request to fetch the product attributes was made.
Without this change, it was impossible to distinguish if the product
has no attributes or if the attributes were not fetched and the client
should then fetch them.

Also, added a fix for the DeliveryBundle type in checkout which
was missing some properties.

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/products/serverInitialState/products.ts
+++ b/packages/redux/src/products/serverInitialState/products.ts
@@ -51,7 +51,11 @@ const serverInitialState: ProductsServerInitialState = ({
     // Send this to the entity's `adaptProductImages`
     productImgQueryParam,
     // Data needed to support product information
-    attributes: productAttributes,
+    // Set attributes to undefined if the productAttributes array
+    // is empty. This is because by default the server render will
+    // return an empty array, even when no request to fetch the
+    // product attributes was made.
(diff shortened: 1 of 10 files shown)


packages/redux/src/products/serverInitialState/products.ts
```diff
@@ -51,7 +51,14 @@ const serverInitialState: ProductsServerInitialState = ({
     // Send this to the entity's `adaptProductImages`
     productImgQueryParam,
     // Data needed to support product information
-    attributes: productAttributes,
+    // Set attributes to undefined if the productAttributes array
+    // is empty. This is because by default the server render will
+    // return an empty array, even when no request to fetch the
+    // product attributes was made.
+    attributes:
+      productAttributes && productAttributes.length === 0
```
nelsonleite, line 59:
```suggestion
      productAttributes?.length === 0
```

Nitpick 😇

nelsonleite (approved):

ruifcnunes (approved):

PR #668, opened 2023-02-16
feat(client|redux): refactor guest orders

andreecosta:
## Description

- This reviews the guest orders redux storage

BREAKING CHANGE:
- Removed getGuestOrder client and fetchGuestOrder action since it works the same way
as getOrder and fetchOrder.

### Dependencies

None.

## Checklist

## Disclaimer


boliveira:
Do not forget to add the `BREAKING CHANGE` label and the `BREAKING CHANGE` footer in the commit and in the PR description for all modules that were removed/changed.

packages/client/src/orders/getGuestOrder.ts
```diff
@@ -1,21 +1,21 @@
...
 /**
- * Method responsible for fetching the details of a guest user order list.
+ * Method responsible for fetching the details of a guest user order.
  *
  * @param config - Custom configurations to send to the client instance (axios).
  *
  * @returns Promise that will resolve when the call to the endpoint finishes.
  */
-const getGuestOrders: GetGuestOrders = config => {
+const getGuestOrder: GetGuestOrder = (orderId, config) => {
```
boliveira, line 12: This new client is added but no matching action is added in redux. I know that this client and the `getOrder` client are both handled by the same code by the backend but we either have 2 choices here:

- Remove this endpoint altogether and use only the `getOrder` and `fetchOrder` to handle it.
- Add a matching `fetchGuestOrder` factory that uses this client but dispatches `FETCH_ORDER_` actions (just like `fetchGuestOrderLegacy` does currently).

I am leaning towards removing this client altogether to avoid confusion with the consumers but discuss this with the team what is the best approach here.

packages/react/src/orders/hooks/useOrders.ts
```diff
@@ -69,19 +67,8 @@ function useOrders(options: UseOrdersOptions = {}) {
...
-    return fetchGuestOrders(fetchConfig);
-  }, [
-    fetchConfig,
-    fetchGuestOrders,
-    fetchQuery,
-    fetchUserOrders,
-    isAuthenticated,
-    userId,
-  ]);
+    return undefined;
```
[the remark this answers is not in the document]
andreecosta: Done 🙂

packages/client/src/orders/getGuestOrder.ts
andreecosta: You are right, we decided to remove it.. I just forgot to delete it 😅
Done!

boliveira (approved):

pedro-gbf (approved):

PR #669, opened 2023-02-16
Feat(client|redux): add set bags promo codes

ivomota:
## Description

This PR adds the ability to set a promo code in the bag (at this moment only one promo code can be applied)

### Dependencies

None.

## Checklist

## Disclaimer


packages/client/src/bags/types/putBagPromocodes.types.ts
```diff
@@ -0,0 +1,25 @@
...
+export type BagPromoCodesInformation = {
+  promoCodesInformation: PromoCodeInformation[];
+};
+
+export type PutBagPromocodesData = {
+  promocodes: string[];
+};
+
+export type PutBagPromocodes = (
+  id: Bag['id'] | null,
```
boliveira, line 22: According to the api spec, this should not be null. Can you confirm, please?

packages/client/src/bags/types/putBagPromocodes.types.ts
```diff
@@ -0,0 +1,25 @@
...
+export type PromoCodeInformation = {
+  promoCode: string;
+  isValid: boolean;
+  error: {
+    code: string;
+    message: string;
+  };
+};
+
+export type BagPromoCodesInformation = {
```
boliveira, line 13: Try to keep the naming consistent, there are some types that refer to `promoCode` (`PromoCodeInformation`, `BagPromoCodesInformation`) while other types are with `promocode` (`PutBagPromocodesData`, `PutBagPromocodes`). In my view, promocode should not be in camel case so this type should be named `BagPromocodesInformation`. Please, review any usages of promocode in the code to have this naming style.

tests/__fixtures__/bags/bag.fixtures.ts
```diff
@@ -47,6 +47,23 @@ export const mockBagItemData = {
   quantity: 99,
 };
 
+export const mockBagPromocodesResponse = {
+  promoCodesInformation: [
+    {
+      promoCode: 'code-1',
+      isValid: true,
+      error: {
```
boliveira, line 55: Is the error set in the response when `isValid` is true? If not, then it is better to have the mock be as realistic as possible with what the endpoint returns. If you need to test a promocode that contains an error, add another element to the array with `isValid` equal to `false` and the error set.

packages/redux/src/bags/actions/__tests__/setBagPromocodes.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    store = bagMockStore(mockState);
+  });
+
+  it('should create the correct actions when the set bag promocodes procedure fails', async () => {
+    const expectedError = new Error('set bag promocodes error');
+
+    (putBagPromocodes as jest.Mock).mockRejectedValueOnce(expectedError);
+
+    expect.assertions(4);
```
boliveira, line 44: This test can be refactored a little bit to avoid this `expect.assertions` call to be necessary which can be error prone like this:

```js
it('should create the correct actions when the set bag promocodes procedure fails', async () => {
    const expectedError = new Error('set bag promocodes errorsss');

    (putBagPromocodes as jest.Mock).mockRejectedValueOnce(expectedError);

    await expect(
      setBagPromocodes(mockBagId, mockBagPromocodesData)(store.dispatch),
    ).rejects.toThrow(expectedError);
    expect(putBagPromocodes).toHaveBeenCalledTimes(1);
    expect(putBagPromocodes).toHaveBeenCalledWith(
      mockBagId,
      mockBagPromocodesData,
      expectedConfig,
    );
    expect(store.getActions()).toEqual(
      expect.arrayContaining([
        {
          type: actionTypes.SET_BAG_PROMOCODES_REQUEST,
        },
        {
          payload: {
            error: expectedError,
          },
          type: actionTypes.SET_BAG_PROMOCODES_FAILURE,
        },
      ]),
    );
  });
```

You can do the same for the remaining test.

packages/redux/src/bags/actionTypes.ts
```diff
@@ -14,6 +14,22 @@ export const ADD_BAG_ITEM_REQUEST =
 export const ADD_BAG_ITEM_SUCCESS =
   '@farfetch/blackout-redux/ADD_BAG_ITEM_SUCCESS';
 
+/**
+ * Action type dispatched when the set promocodes from bag request fails.
```
boliveira, line 18: I think this reads better:

```suggestion
 * Action type dispatched when the set bag promocodes request fails.
```

What do you think? If you agree, then do the same for the remaining actions, please.

packages/redux/src/bags/actionTypes.ts
```diff
@@ -14,6 +14,22 @@ export const ADD_BAG_ITEM_REQUEST =
...
+ */
+export const SET_BAG_PROMOCODES_FAILURE =
+  '@farfetch/blackout-core/SET_BAG_PROMOCODES_FAILURE';
```
boliveira, line 21:
```suggestion
  '@farfetch/blackout-redux/SET_BAG_PROMOCODES_FAILURE';
```
Same for the remaining actions.

packages/redux/src/bags/actions/factories/setBagPromocodesFactory.ts
```diff
@@ -0,0 +1,52 @@
...
+  PutBagPromocodes,
+  PutBagPromocodesData,
+  toBlackoutError,
+} from '@farfetch/blackout-client';
+import type { Dispatch } from 'redux';
+
+/**
+ * Creates a thunk factory configured with the specified client to set.
+ *
+ * @param patchBagItem - Patch bag item client.
```
boliveira, line 15:
```suggestion
 * @param putBagPromocodes - Put bag promocodes client.
```

packages/redux/src/bags/actions/setBagPromocodes.ts
```diff
@@ -0,0 +1,8 @@
+import { putBagPromocodes } from '@farfetch/blackout-client';
+import { setBagPromocodesFactory } from './factories';
+
+/**
+ * Updates a specific bag item with the given params.
```
boliveira, line 5:
```suggestion
 * Sets the passed promocodes to the bag.
```

packages/redux/src/bags/actions/setBagPromocodes.ts
```diff
@@ -0,0 +1,8 @@
...
+ */
+
```
boliveira, line 7: This empty line can be removed

packages/redux/src/bags/reducer.ts
```diff
@@ -190,6 +195,12 @@ const bagOperations = (
   }
 };
 
+const bagPromocodes = reducerFactory(
+  'SET_BAG_PROMOCODES',
+  INITIAL_STATE.bagPromocodes,
+  actionTypes,
+);
+
 export const entitiesMapper = {
```
boliveira, line 204: Take a look at the entitiesMapper for the `LOGOUT_SUCCESS` action which is not resetting the `bagPromocodesInformation` entities (and the `bagOperations` entities as well which I believe it should). Perhaps it would be best to extract the reset logic to a common function and use that function in `RESET_BAG_ENTITIES` and `LOGOUT_SUCCESS` since they should do the same.

packages/redux/src/bags/selectors.ts
```diff
@@ -711,3 +711,27 @@ export const isBagOperationLoading = (
...
+/**
+ * Retrieves current promocodes information in bag.
+ *
+ * @param state - Application state.
+ *
+ * @returns - Bag PromoCodes Information, `undefined` if there are no bag id
+ *
+ * @example
+ * ```
+ * import { getBagPromoCodesInformation } from '@farfetch/blackout-core/bags/redux';
```
boliveira, line 724: I have these changes in another PR but all examples should use the import from the root of the package like this:

```suggestion
 * import { getBagPromoCodesInformation } from '@farfetch/blackout-core';
```

packages/redux/src/bags/selectors.ts
```diff
@@ -711,3 +711,27 @@ export const isBagOperationLoading = (
...
+ *
+ * const mapStateToProps = state => ({
+ *     bagPromoCodesInformation: getBagPromoCodesInformation(state)
+ * });
+ * ```
+ */
+export const getBagPromoCodesInformation = (state: StoreState) => {
+  const bagId = getBagId(state);
+  if (!bagId) {
```
boliveira, line 733: Add some line spacing here to make it easier to read, please.

packages/redux/src/bags/reducer.ts
```diff
@@ -1,6 +1,7 @@
 import * as actionTypes from './actionTypes';
```
boliveira, line 1: I do not see selectors `areBagPromocodesLoading` and `getBagPromocodesError`. Aren't they necessary as well?

packages/client/src/bags/types/putBagPromocodes.types.ts
ivomota: I will change this but the putBagPromocodes response from the API will give us an object with `promoCodesInformation`

boliveira: Yes, sorry, only change the naming where it is safe to do so. If the endpoint returns `promoCodesInformation`, we cannot change that specific name but everywhere else where it does not refer to names shared with the server, we can.

packages/redux/src/bags/selectors.ts
tiagodicosta: I think it is `@farfetch/blackout-redux`, or am I wrong?

packages/redux/src/bags/actions/setBagPromocodes.ts
ivomota: I will change that to be similar to the API definition:
`Sets the list of promocodes to be applied to the bag.`

packages/redux/src/bags/reducer.ts
```diff
@@ -238,8 +250,13 @@ export const getIsBagOperationLoading = (state: BagsState) =>
   state.bagOperations.isLoading;
 export const getBagOperationError = (state: BagsState) =>
   state.bagOperations.error;
+export const getIsBagPromocodesLoading = (state: BagsState) =>
```
tiagodicosta, line 253:
```suggestion
export const areBagPromocodesLoading = (state: BagsState) => state.bagPromocodes.isLoading;
export const getBagPromocodesError = (state: BagsState) => state.bagPromocodes.error;
```

packages/redux/src/bags/reducer.ts
```diff
@@ -238,8 +250,13 @@ export const getIsBagOperationLoading = (state: BagsState) =>
...
+  state.bagOperations.isLoading;
+export const getBagPromocodesError = (state: BagsState) =>
```
tiagodicosta, line 255:
```suggestion
export const getBagPromocodesError = (state: BagsState) => state.bagPromocodes.error;
```

packages/client/src/bags/__fixtures__/putBagPromocodes.fixtures.ts
```diff
@@ -0,0 +1,17 @@
+import { rest, RestHandler } from 'msw';
+import type { BagPromoCodesInformation } from '../types';
+
+const path = '/api/commerce/v1/bags/:bagId/promocodes';
+
+const fixtures = {
+  success: (response: BagPromoCodesInformation): RestHandler =>
+    rest.put(path, async (_req, res, ctx) =>
```
tiagodicosta, line 8: this async and on failure are unnecessary.

packages/client/src/bags/__tests__/putBagPromocodes.test.ts
```diff
@@ -0,0 +1,44 @@
+import {
+  mockBagId,
+  mockBagPromocodesData,
+  mockBagPromocodesResponse,
```
tiagodicosta, line 4: In my point of view, we can create a new fixtures file for promocodes and separate them from the bag.

packages/client/src/bags/types/putBagPromocodes.types.ts
ivomota: You are right. Fixed

packages/client/src/bags/types/putBagPromocodes.types.ts
ivomota: Changed 👍

tests/__fixtures__/bags/bag.fixtures.ts
ivomota: Yes, it makes sense. I removed the error object for a valid mock

packages/redux/src/bags/actionTypes.ts
ivomota: Changed 👍

packages/redux/src/bags/actionTypes.ts
ivomota: Fixed 👍

packages/redux/src/bags/actions/factories/setBagPromocodesFactory.ts
ivomota: Fixed 👍

packages/redux/src/bags/actions/setBagPromocodes.ts
ivomota: Changed 👍

packages/redux/src/bags/actions/setBagPromocodes.ts
ivomota: Done

packages/redux/src/bags/reducer.ts
ivomota: I created a function called `resetBagEntities` and I made something similar to what we do on the checkout reducer.
Please review this.

packages/redux/src/bags/actions/__tests__/setBagPromocodes.test.ts
ivomota: It looks very good to me.
I changed the 2 tests.

packages/redux/src/bags/selectors.ts
ivomota: You are right @tiagodicosta 
Fixed 👍

packages/redux/src/bags/selectors.ts
ivomota: done

packages/redux/src/bags/reducer.ts
ivomota: This was a mistake. Fixed 👍

packages/redux/src/bags/reducer.ts
ivomota: Fixed 👍

packages/redux/src/bags/reducer.ts
ivomota: Fixed 👍

packages/client/src/bags/__fixtures__/putBagPromocodes.fixtures.ts
ivomota: Fixed 👍

packages/client/src/bags/__tests__/putBagPromocodes.test.ts
ivomota: I thought about that too but promo codes besides bag items and bag operation are something small and that depends on the bag so I think it might be better to be inside the bag fixtures. Is it ok for you?

packages/client/src/bags/__fixtures__/putBagPromocodes.fixtures.ts
tiagodicosta: the async is still in the successful case.

ivomota: 😓 fixed now

packages/client/src/bags/__tests__/putBagPromocodes.test.ts
tiagodicosta: ok, fine!

tiagodicosta (approved):

packages/client/src/bags/types/putBagPromocodes.types.ts
boliveira: In `packages/redux/src/bags/actions/factories/setBagPromocodesFactory.ts`, the entities name can be renamed to `bagPromocodesInformation` as well to keep consistency. The remaining ones, are correctly renamed, thanks!

packages/redux/src/bags/actions/factories/setBagPromocodesFactory.ts
boliveira: The name of the param is still `patchBagItem`, can you change it, please?

packages/redux/src/bags/actions/setBagPromocodes.ts
boliveira: Much better than my suggestion, thanks!

packages/redux/src/bags/reducer.ts
boliveira: That is great, thanks!

packages/redux/src/bags/reducer.ts
boliveira: I still do not see those selectors, have you pushed those changes?

packages/client/src/bags/types/putBagPromocodes.types.ts
ivomota: Fixed now @boliveira

packages/redux/src/bags/actions/factories/setBagPromocodesFactory.ts
ivomota: You are right! Fixed

boliveira (approved):

PR #673, opened 2023-02-17
fix(redux): checkout reducer not overwriting address state

anapim:
## Description

This fixes an issue where the properties `state.code` and `state.name` in the shipping/billing address of a checkout order would not be cleared in the store if their value was `undefined`.

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/checkout/__tests__/reducer.test.ts
+++ b/packages/redux/src/checkout/__tests__/reducer.test.ts
@@ -11,6 +11,8 @@ import {
   mockCheckoutDetailsEntity,
   mockCheckoutOrderItemEntity,
   mockCheckoutOrderItemProductsEntity,
+  mockCheckoutState,
+  mockCollectPoint,
   mockCollectPointsResponse,
   mockDeliveryBundlesResponse,
   mockGetOperationActionPayload,
@@ -1294,6 +1296,7 @@ describe('checkout reducer', () => {
       const result = {
         checkoutOrders: {
           1: {
+            ...mockCheckoutState.entities.checkoutOrders[checkoutOrderId],
             checkout: 123,
             tags: ['NEW_GIFT'],
           },
@@ -1302,13 +1305,14 @@ describe('checkout reducer', () => {
       const state = {
         checkoutOrders: {
           1: {
+            ...mockCheckoutState.entities.checkoutOrders[checkoutOrderId],
             checkout: 123,
             tags: ['GIFT'],
           },
         },
       };
 
-      it.each([
+      it.each<keyof typeof entitiesMapper>([
         actionTypes.CREATE_CHECKOUT_ORDER_SUCCESS,
         actionTypes.SET_CHECKOUT_ORDER_PROMOCODE_SUCCESS,
         actionTypes.FETCH_CHECKOUT_ORDER_SUCCESS,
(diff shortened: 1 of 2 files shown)


packages/redux/src/checkout/reducer.ts
```diff
@@ -172,20 +172,70 @@ const convertCheckoutOrder = (
...
+      !!get(
+        mergedState,
+        `checkoutOrders[${result}].billingAddress.state.name`,
+      )) &&
+    (!get(entities, `checkoutOrders[${result}].billingAddress.state.code`) ||
+      !get(entities, `checkoutOrders[${result}].billingAddress.state.name`));
+
+  let returnedState = mergedState;
+
+  if (shouldResetShippingState || shouldResetBillingState) {
```
boliveira, line 201: This will replace both addresses if only one of them has changed, is that the intention here?

packages/redux/src/checkout/reducer.ts
```diff
@@ -172,20 +172,70 @@ const convertCheckoutOrder = (
...
+    returnedState = {
+      ...returnedState,
```
boliveira, line 203: As stated before, using immer will make these updates a lot easier and less error prone.

packages/redux/src/checkout/__tests__/reducer.test.ts
```diff
@@ -1326,6 +1326,80 @@ describe('checkout reducer', () => {
...
+              },
+              billingAddress: {
+                state: {},
+              },
+            },
+          },
+        };
+
+        expect(
+          // @ts-expect-error
```
boliveira, line 1364: Why was this `ts-expect-error` necessary? If there is a valid reason, add the reason in front of it as it let others know why was it necessary without removing the comment.

packages/redux/src/checkout/reducer.ts
```diff
@@ -172,20 +172,70 @@ const convertCheckoutOrder = (
     !!get(mergedState, `checkoutOrders[${result}].clickAndCollect`) &&
     !get(entities, `checkoutOrders[${result}].clickAndCollect`);
 
-  if (!shouldResetClickAndCollect) {
-    return mergedState;
+  const shouldResetShippingState =
+    (!!get(
```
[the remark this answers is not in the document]
anapim: I followed your suggestions, can you review again?

packages/redux/src/checkout/reducer.ts
```diff
@@ -172,20 +173,72 @@ const convertCheckoutOrder = (
...
+        collectpoints: draftState.checkoutOrders[result]?.collectpoints,
         ...entities?.checkoutOrders[result],
-      },
-    },
-  };
+      };
+    }
+
+    if (shouldReplaceShippingState) {
+      draftState.checkoutOrders[result] = {
```
boliveira, line 223: It is not necessary to reassign the checkout order in draftState since immer will take care of that.
In this case what you wanted is to do something like this:

```js
draftState.checkoutOrders[result]?.shippingAddress.state = entities.checkoutOrders[result]?.shippingAddress?.state;
```

Then, when the call to `produce` finishes, immer will check which properties have changed and will generate a new state object with new references only for those properties that changed.

If, instead, you meant to replace the whole shipping address if `state` has changed, then you will need to adapt to something like this:

```js
draftState.checkoutOrders[result]?.shippingAddress= entities.checkoutOrders[result]?.shippingAddress;
```
So, in summary, you only need to treat the `draftState` as a mutable object and immer will take care of generating the object will new references where necessary.

andrevechina (approved):

boliveira (approved):

edmundofreitas (approved):

nelsonleite (approved):

PR #678, opened 2023-02-22
feat(redux) split orders - fetch order and return options

pedro-gbf:
#### Fetch Order changes
This removes the aggregation of order details and order summaries and makes the appropriate changes on the selectors. 
*  Removed the entities mapper specific fetch order logic to merge information since we no longer have merging issues.
* getOrder was reviewed
* getOrderReturns was migrated and reviewed on the returns area.
* getOrderMerchants was deleted and a new selector was born to replace it called `getOrderSummaries` that returns all the order summaries of a certain order id.
* getOrderItemsByMerchant was renamed to getOrderItemsBySummary and it now returns the order items of an order summary instead of splitting them by merchant.
* getOrderItemQuantity was reviewed to remove an unnecessary restriction
* getUserOrder is now normalising the date
* getOrderItems is now normalising the result
* getOrderItem is now normalising the result
* Tests were adapted and added to test the mentioned changes

#### Fetch Order Return Options changes
* Removed the reset of the return options since they are already resetted when a new user is fetches or when a user logins. There is no use case to manually reset this data for a user.
* Revamped the returnOption schema to wrap around the entire object that contains sub returnOptions per type. The older schematisation implied that there was a need to reference each of the sub returnOption and there is none. Also the identifier of the this entity has been changed to be the `merchantOrderId` property so we can link it to the order items if needed.
* Removed the artificial merchant normalization.
* getOrderReturnOptions selector now returns a list of return options denormalized organised by `merchantOrderId` this id can be linked to order items.
* getReturnOption  is now normalising the result. You now need to provide a `merchantOrderId` to fetch it.
* fetchOrderReturnOptions now saves the array with the merchantOrderIds identifying the returnOptions entities in the Orders area of the redux store.

## Checklist

## Disclaimer


packages/redux/src/entities/types/returnOptions.types.ts
```diff
@@ -1,25 +1,9 @@
...
-};
+export type ReturnOptionEntity = MerchantOrderReturnOptions;
 
-export type ReturnOptionEntityDenormalized = Omit<
-  ReturnOptionEntity,
-  'merchant'
-> & {
-  merchant: MerchantEntity;
-};
+export type ReturnOptionEntityDenormalized = Array<ReturnOptionEntity>;
```
boliveira, line 5: This type does not seem right and it is not being used anywhere. Can you take a look at it?

packages/redux/src/orders/selectors.ts
```diff
@@ -292,88 +333,94 @@ export const getOrder: (
...
+    }
+
+    if (!returnOptions) {
+      return undefined;
+    }
+
+    const result: Record<
+      MerchantOrderReturnOptions['merchantOrderId'],
+      MerchantOrderReturnOptions
+    > = {};
```
boliveira, line 377: There is a problem with this return value. When there are no return options entities, this selector will return undefined, which is good, because the caller can then infer that no request was made to fetch the return options and then he can make the request. The problem is when you invoke this selector with another orderId without removing the previous order return options entities. It will return an empty object, but no request was made to the server to fetch the order return options, so we will end up returning an invalid response. We need to change this to sync with the result of the fetch order return options action, which will need changes to the reducer to store the result as well to have it work similarly to the way the `getUserOrdersResult` works.

nelsonleite (changes requested):

packages/redux/src/orders/selectors.ts
```diff
@@ -272,7 +313,7 @@ export const getReturnOptions = (state: StoreState) =>
  *
  * @returns Return option object.
  */
-export const getReturnOption = (state: StoreState, returnOptionId: string) =>
+export const getReturnOption = (state: StoreState, returnOptionId: number) =>
```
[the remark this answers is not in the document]
pedro-gbf: Fixed, changed the name of the variable as well, nice catch.

packages/redux/src/orders/selectors.ts
pedro-gbf: Done, please validate the solution 👍

packages/redux/src/entities/types/returnOptions.types.ts
pedro-gbf: Yep it was wrong, it was now fixed, please validate.

nelsonleite (approved):

boliveira (approved):

PR #685, opened 2023-02-27
feat(client): add metadata to bag item and requests types

anapim:
## Description

This adds the property `metadata` to bag item and respective requests types.

## Checklist

## Disclaimer

### DIFF
--- a/packages/client/src/types/common/metadata.types.ts
+++ b/packages/client/src/types/common/metadata.types.ts
@@ -1,0 +1,5 @@
+import type { JSONValue } from './jsonValue.types';
+
+export type Metadata = {
+  [k: string]: JSONValue;
+};
--- a/packages/client/src/bags/types/bagItem.types.ts
+++ b/packages/client/src/bags/types/bagItem.types.ts
@@ -6,8 +6,11 @@ import type {
   ProductImageGroup,
 } from '../../types';
 import type { Brand } from '../../brands/types';
+import type { Metadata } from '../../types/common/metadata.types';
 import type { Price, Product, ProductVariant, Size } from '../../products';
 import type { ProductCategory } from '../../categories/types';
+
+export type BagItemMetadata = Metadata;
 
 export type BagItem = {
   id: number;
@@ -52,4 +55,5 @@ export type BagItem = {
     images: ProductImageGroup;
   } | null;
   gender: GenderCode;
+  metadata?: BagItemMetadata;
 };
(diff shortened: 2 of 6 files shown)


edmundofreitas (approved):

packages/client/src/types/common/metadata.types.ts
```diff
@@ -0,0 +1,3 @@
+export type Metadata = {
+  [k: string]: unknown;
```
boliveira, line 2: Can we be more explicit instead of using `unknown`? For example, with this definition, I could pass functions and objects, which should not be allowed. So perhaps this could be defined as:

```suggestion
  [k: string]: number | string | boolean;
```

I think these types are the ones that make the most sense to be used for metadata. What do you think?

edmundofreitas: Well accordingly to the apispec, in fact this is an object that accepts anything that is supported in json.

edmundofreitas: so you could possible add inside this, another objects, arrays, etc. So I agree to keep unknown here 🤔

nelsonleite (approved):

packages/client/src/types/common/metadata.types.ts
boliveira: I see. I agree regarding objects and arrays but functions I think should not be allowed. Let me check if there is already a type that we can reuse for that.

boliveira: There is not one type readily available and there is a lot of [discussion](https://github.com/microsoft/TypeScript/issues/1897) about it but I tried this type definition and it worked:

```js

export type JSONValue =
  | null
  | boolean
  | number
  | string
  | JSONValue[]
  | {[key: string]: JSONValue};

export type Metadata = {
  [k: string]: JSONValue;
}
```

What do you think? If you adopt this solution do not forget to export the type `JSONValue` from the package as well (not the Metadata type as it is only an alias) and not only from the file.

anapim: I like this approach but I separated `JSONValue` to a new file, because I believe in can be used in other situations and it would makes more sense on its own.

edmundofreitas: I used that type before, it works for me :)

boliveira (approved):

packages/client/src/types/common/metadata.types.ts
boliveira: Thank you!

PR #691, opened 2023-03-08
feat(analytics|react): add unique id to GTM integration for next

danielbento92:
## Description

- Added new property on track events, to set __blackoutAnalyticsEventId
on GTM. This property is already used on other integrations to relation
events on different integrations.

- Code refactor: created new contant on analytics package to define
default property name for __blackoutAnalyticsEventId, avoiding possible inconsistencies.

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/analytics/src/utils/constants.ts
+++ b/packages/analytics/src/utils/constants.ts
@@ -17,7 +17,8 @@ export const CONSENT_KEYS = [
   DefaultConsentKeys.MARKETING,
   DefaultConsentKeys.PREFERENCES,
 ];
+
 export const CONSENT_CATEGORIES_PROPERTY = 'consentCategories';
 export const LOAD_INTEGRATION_TRACK_TYPE = 'loadIntegration';
 export const ON_SET_USER_TRACK_TYPE = 'onSetUser';
-export const ANALYTICS_UNIQUE_EVENT_ID = '__uniqueEventId';
+export const ANALYTICS_UNIQUE_EVENT_ID = '__blackoutAnalyticsEventId';
--- a/packages/analytics/src/__tests__/__snapshots__/index.test.ts.snap
+++ b/packages/analytics/src/__tests__/__snapshots__/index.test.ts.snap
@@ -133,7 +133,7 @@ Object {
     "Omnitracking": [Function],
   },
   "utils": Object {
-    "ANALYTICS_UNIQUE_EVENT_ID": "__uniqueEventId",
+    "ANALYTICS_UNIQUE_EVENT_ID": "__blackoutAnalyticsEventId",
     "CONSENT_CATEGORIES_PROPERTY": "consentCategories",
     "CONSENT_KEYS": Array [
       "statistics",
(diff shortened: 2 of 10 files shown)


ruifcnunes (approved):

packages/analytics/src/utils/constants.ts
```diff
@@ -17,7 +17,10 @@ export const CONSENT_KEYS = [
   DefaultConsentKeys.MARKETING,
   DefaultConsentKeys.PREFERENCES,
 ];
+
 export const CONSENT_CATEGORIES_PROPERTY = 'consentCategories';
 export const LOAD_INTEGRATION_TRACK_TYPE = 'loadIntegration';
 export const ON_SET_USER_TRACK_TYPE = 'onSetUser';
 export const ANALYTICS_UNIQUE_EVENT_ID = '__uniqueEventId';
```
boliveira, line 24: I think this property has the same purpose as the new one just created. Can you check, please? If they are supposed to do the same thing, then delete this one.

packages/analytics/src/utils/constants.ts
```diff
@@ -17,7 +17,10 @@ export const CONSENT_KEYS = [
...
+export const ANALYTICS_UNIQUE_EVENT_INTEGRATION_DEFAULT_NAME =
```
boliveira, line 25: This name I think it is not correct since this is not a name for the integration but for the uniqueEventId (As I said before, I think the property `ANALYTICS_UNIQUE_EVENT_ID ` refers to the same thing)

boliveira: After looking at the code more thoroughly, I see that this prop is just a rename for the default `__uniqueEventId`. The question I have is, why do we have this indirection? Couldn't we have this instead?

```suggestion
export const ANALYTICS_UNIQUE_EVENT_ID = '__blackoutAnalyticsEventId';
```
Then we just need to use the `ANALYTICS_UNIQUE_EVENT_ID` instead of using `ANALYTICS_UNIQUE_EVENT_INTEGRATION_DEFAULT_NAME `.

danielbento92: This __blackoutAnalyticsEventId, appeared for the first time in GA4, and now it was replicated into GTM integration, always mapping from analytics __uniqueEventId field.

As these two integrations use the same property name on their datalayer, I set this const on our analytics package, to represent a default name in this kind of situation.

boliveira: I understand that, I am just highlighting the fact that we do not need 2 variables for this job. We could just change the value of the `ANALYTICS_UNIQUE_EVENT_ID` to `__blackoutAnalyticsEventId` and use only that value throughout all integrations. If another integration needs that value to be sent in another property (like Omnitracking does), it can remap it there to the name of its choosing. By following this suggestion you can then change both the GA4 and GTM integrations code to be like this:

```js
const payload = {
      ...(eventProperties as object),
      [coreUtils.ANALYTICS_UNIQUE_EVENT_ID]: get(
      data,
      `context.event.${coreUtils.ANALYTICS_UNIQUE_EVENT_ID}`,
    ),
      type: data.type,
      event: data.event,
    };

```

And we remove the redundancy.

danielbento92: I agree that could be more simple and I've applied your suggestion.

boliveira (approved):

PR #694, opened 2023-03-09
feat(client): add get return workflow endpoint

jpklzm:
## Description

This adds getReturnWorkflow endpoint.

## Checklist

## Disclaimer


nelsonleite (approved):

packages/client/src/returns/types/getReturnWorkflow.types.ts
```diff
@@ -0,0 +1,14 @@
+import type { Config } from '../..';
+import type { Return } from './return.types';
+
+export type ReturnWorkflow = {
+  isCompleted: boolean;
+  sort: number;
+  status: { code: string };
```
tiagodicosta, line 7: I think we can have an enum for the status code instead of assuming just **string**.
```suggestion
  status: { code: StatusCodeReturnWorkflows };
```

```
enum StatusCodeReturnWorkflows {
    WAITING_EXTERNAL_APPROVAL = 'WaitingExternalApproval',
    DRAFT = 'Draft',
    BOOKED = 'Booked',
    AT_PARTNER_LOCATION = 'AtPartnerLocation',
    IN_TRANSIT = 'InTransit',
    ARRIVED_AT_FINAL_LOCATION = 'ArrivedAtFinalLocation',
    PROCESSING = 'Processing',
    ACCEPTED = 'Accepted',
    PARTIALLY_ACCEPTED = 'PartiallyAccepted',
    REFUSED = 'Refused',
    CANCELLED = 'Cancelled',
};
```

boliveira: Just to add to tiago's suggestion, yes we should if possible add an enum here and in that case, the name of the enum items should be in pascal case, i.e.:

```js
enum StatusCodeReturnWorkflows {
    WaitingExternalApproval = 'WaitingExternalApproval',
    ...
```

packages/client/src/returns/types/getReturnWorkflow.types.ts
```diff
@@ -0,0 +1,14 @@
+import type { Config } from '../..';
+import type { Return } from './return.types';
+
+export type ReturnWorkflow = {
```
boliveira, line 4: Could we define this type instead like this?

```js
export type ReturnWorkflowStep = {
  isCompleted: boolean;
  sort: number;
  status: { code: string };
  statusDateUTC?: string;
};

export type ReturnWorkflow = ReturnWorkflowStep[];
```

Let me know if the name `ReturnWorkflowStep` makes sense as that is what I figured those elements inside the array represent but I can be wrong :)

boliveira (commented):
Some little changes are needed but overall it is good to me!

tiagodicosta (approved):

packages/client/src/returns/types/getReturnWorkflow.types.ts
```diff
@@ -0,0 +1,30 @@
+import type { Config } from '../..';
+import type { Return } from './return.types';
+
+export enum ReturnWorkflowStatuses {
```
boliveira, line 4: Since this enum is not used as a bitfield or flags, the name should be in singular

```suggestion
export enum ReturnWorkflowStatus {
```

boliveira (approved):

PR #717, opened 2023-04-04
feat(core): exclude update product event

danielbento92:
## Description

- Adjust middlewares:
  - Changing size on cart or wishlist pages will trigger only change size
   - Changing quantity always add/remove products events and change quantity event
    - When add product that was already on bag, should not trigger change quantity event.
- Add tests

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/analytics/redux/middlewares/__tests__/bag.test.js
+++ b/packages/core/src/analytics/redux/middlewares/__tests__/bag.test.js
@@ -3,7 +3,7 @@ import { bagMiddleware } from '../../';
 import { combineReducers } from 'redux';
 import { logger } from '../../../utils';
 import { mockStore } from '../../../../../tests';
-import Analytics, { eventTypes } from '../../../';
+import Analytics, { eventTypes, fromParameterTypes } from '../../../';
 import merge from 'lodash/merge';
 
 // Mock logger so it does not output to the console
@@ -446,11 +446,11 @@ describe('bagMiddleware()', () => {
       });
     });
 
-    it('Should track a remove and an add event if the product sizes changed', async () => {
+    it('should trigger only PRODUCT_UPDATED event when it changed size event, with sizeId property assigned on event payload.', async () => {
       const simplifiedStore = mockSimplifiedStore(baseMockState, [
         bagMiddleware(analytics),
       ]);
-
+      const size = { id: 999, name: 'XXL' };
       await simplifiedStore.dispatch({
         type: bagActionTypes.UPDATE_BAG_ITEM_SUCCESS,
         payload: {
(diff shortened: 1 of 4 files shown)


packages/core/src/analytics/redux/middlewares/__tests__/bag.test.js
```diff
@@ -492,23 +493,119 @@ describe('bagMiddleware()', () => {
...
+      expect(trackSpy).nthCalledWith(
+        2,
+        eventTypes.PRODUCT_ADDED_TO_CART,
+        expect.objectContaining({
+          quantity: 1,
+        }),
+      );
+    });
+
+    it('should not trigger any change quantity, only add to bag when it changed quantity by PDP', async () => {
```
ruifcnunes, line 578:
```suggestion
    it('should trigger PRODUCT_ADDED_TO_CART when quantity changes from PDP', async () => {
```

packages/core/src/analytics/redux/middlewares/__tests__/bag.test.js
```diff
@@ -492,23 +493,119 @@ describe('bagMiddleware()', () => {
...
+          },
+        },
+        meta: {
+          productId,
+          quantity: baseMockState.entities.bagItems[bagItemId].quantity + 1,
+          from: fromParameterTypes.PDP,
+        },
+      });
+
+      // expect trigger analytics product updated event
```
ruifcnunes, line 601:
```suggestion
      // expect analytics PRODUCT_ADDED_TO_CART event to be triggered
```

packages/core/src/analytics/redux/middlewares/__tests__/wishlist.test.js
```diff
@@ -724,6 +776,130 @@ describe('wishlistMiddleware()', () => {
         wishlistId,
       });
     });
+
+    it('should trigger only change size event when it changed size of product on wishlist', async () => {
```
ruifcnunes, line 780:
```suggestion
    it('should trigger PRODUCT_UPDATED event when the product size is updated on wishlist', async () => {
```

packages/core/src/analytics/redux/middlewares/__tests__/wishlist.test.js
```diff
@@ -724,6 +776,130 @@ describe('wishlistMiddleware()', () => {
...
+
+      // expect trigger analytics product updated event
+      expect(trackSpy).nthCalledWith(
+        1,
+        eventTypes.PRODUCT_UPDATED,
+        expect.objectContaining({ sizeId: size.id }),
+      );
+    });
+
+    it('should trigger change quantity and add to wishlist event when quantity of product is incremented by wishlist page', async () => {
```
ruifcnunes, line 818:
```suggestion
    it('should trigger PRODUCT_UPDATED and PRODUCT_ADDED_TO_WISHLIST when product quantity is incremented on wishlist', async () => {
```

packages/core/src/analytics/redux/middlewares/__tests__/wishlist.test.js
```diff
@@ -724,6 +776,130 @@ describe('wishlistMiddleware()', () => {
...
+      expect(trackSpy).nthCalledWith(
+        3,
+        eventTypes.PRODUCT_ADDED_TO_WISHLIST,
+        expect.objectContaining({
+          quantity: 1,
+        }),
+      );
+    });
+
+    it('should trigger add to wishlist event when quantity of product is incremented by product PDP page', async () => {
```
ruifcnunes, line 866:
```suggestion
    it('should trigger PRODUCT_ADDED_TO_WISHLIST event when a product quantity is incremented on a PDP', async () => {
```

packages/core/src/analytics/redux/middlewares/__tests__/bag.test.js
```diff
@@ -492,23 +493,119 @@ describe('bagMiddleware()', () => {
         quantity: 5,
         size: sizes[1].name,
         sizeId: sizes[1].id,
+        from: fromParameterTypes.BAG,
       };
 
       expect(trackSpy).nthCalledWith(1, eventTypes.PRODUCT_UPDATED, {
```
ruifcnunes, line 499:
```suggestion
      expect(trackSpy).nthCalledWith(1, eventTypes.PRODUCT_UPDATED, baseData);
```

packages/core/src/analytics/redux/middlewares/bag.js
```diff
@@ -278,34 +277,31 @@ export default (analyticsInstance, customActionTypes) => {
...
-          removedProductData.size = data.oldSize;
-          removedProductData.quantity = data.oldQuantity;
-
-          analyticsInstance.track(
-            eventTypes.PRODUCT_REMOVED_FROM_CART,
-            removedProductData,
-          );
 
-          data.oldQuantity = 0;
+        // product updated will only triggered if provenience is from BAG
```
ruifcnunes, line 281:
```suggestion
        // product updated will only be triggered if provenience is from BAG
```

packages/core/src/analytics/redux/middlewares/bag.js
```diff
@@ -278,34 +277,31 @@ export default (analyticsInstance, customActionTypes) => {
...
+        if (data.from === fromParameterTypes.BAG) {
+          // Track ga4 update event
```
ruifcnunes, line 283: (this can be removed - product updated if for any integration that has interest on the event)
```suggestion
```

packages/core/src/analytics/redux/middlewares/bag.js
```diff
@@ -278,34 +277,31 @@ export default (analyticsInstance, customActionTypes) => {
...
         }
 
         data.quantity = Math.abs(data.quantity - (data.oldQuantity || 0));
-        analyticsInstance.track(eventType, { ...data, oldQuantity: undefined });
+
+        // If eventType was set on previous ifs, then track
+        if (eventType !== null) {
+          analyticsInstance.track(eventType, {
+            ...data,
+            oldQuantity: undefined,
```
ruifcnunes, line 302: Should we do this here? 
With this, we're not passing the value to integrations that might have interest in this property. If GA4 does not need `oldQuantity` in these events, then it should be GA4's responsibility to ignore the property, instead of forcing an `undefined` on the analytics track layer.

packages/core/src/analytics/redux/middlewares/wishlist.js
```diff
@@ -261,15 +263,33 @@ export default (analyticsInstance, customActionTypes) => {
...
+          eventType = eventTypes.PRODUCT_ADDED_TO_WISHLIST;
+        }
+
+        data.quantity = Math.abs(data.quantity - (data.oldQuantity || 0));
+
+        // If eventType was set on previous ifs, then track
+        if (eventType !== null) {
+          analyticsInstance.track(eventType, {
+            ...data,
+            oldQuantity: undefined,
```
ruifcnunes, line 290: same as bag middleware

packages/core/src/analytics/redux/middlewares/bag.js
danielbento92: On eventType Product added/removed from cart (and wishlist too) oldQuantity isn't required for these events, so it could be sent any value to track. But in this case, I think we should send undefined because for me, when you put the new quantity value with previous calculation, that result on quantity added or removed of some product (not real quantity), then use the real oldQuantity value seems wrong information.
EG: when you have an product on bag with quantity 10, then you remove 2, will be triggered PRODUCT_REMOVED_FROM_CART, with quantity 2, instead of 8. If we introduce oldQuantity value on analytics, it could induce that you have 10 products, then you remove products and now the quantity it's 2. Beside that we have other events with similar payloads, that quantity and oldQuantity represents different things, such quantity here it's the changedQuantity, but in other events, quantity represents bag or wishlist product quantity.
Other relevant information it's that we already did this(send undefined) on previous code, the only thing that changed it's the if around.

What you think of maintain this here in main the undefined, and change in next to append one field to payload?
- changedQuantity: moved quantity (actual quantity on this events)
- quantity: remove calculation, then use the real value of quantity.
- oldQuantity: send old quantity values.

And in each integration, on this method, map from changedQuantity instead of map from quantity.
With this payload, I think would be more compreensive, because it could be more clear, because in every event, we could know that quantity always represents same thing on each event, instead of what we have now (in some events, quantity represents quantity of items, then in other events, it represents quantity of changed items).

ruifcnunes (approved):

PR #726, opened 2023-04-13
feat(analytics|react): add package version to analytics next

danielbento92:
## Description

- Add parameter to each integration to fill package versions on omni and ga4 requests.
- Omnitracking, are missing update contract that would allow to send this new parameter on final request.

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/analytics/analytics.ts
+++ b/packages/react/src/analytics/analytics.ts
@@ -1,11 +1,21 @@
+import { get } from 'lodash-es';
 import Analytics, {
   TrackTypes as analyticsTrackTypes,
+  type ContextData,
   type EventContextData,
+  type EventData,
   type EventProperties,
   type IntegrationRuntimeData,
   PlatformTypes,
+  type TrackTypesValues,
 } from '@farfetch/blackout-analytics';
 import webContext from './context.js';
+
+const {
+  name: PACKAGE_NAME,
+  version: PACKAGE_VERSION,
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+} = require('../../package.json');
 
 /**
  * Analytics facade for web applications. Refer to \@farfetch/blackout-analytics
@@ -44,7 +54,7 @@ class AnalyticsWeb extends Analytics {
     if (this.currentPageCallData) {
       const { event, properties } = this.currentPageCallData;
 
-      const pageEventData = await super.getTrackEventData(
+      const pageEventData = await this.getTrackEventData(
         analyticsTrackTypes.PAGE,
         event,
         properties,
@@ -53,6 +63,59 @@ class AnalyticsWeb extends Analytics {
       super.forEachIntegrationSafe(loadedIntegrations, integration =>
         integration.track(pageEventData),
       );
+    }
+  }
+
+  /**
+   * Gets event data for a track event.
+   *
+   * @param type         - Type of the event being called.
+   * @param event        - Name of the event from analytics.track call.
+   * @param properties   - Event properties from analytics.track call.
+   * @param eventContext - Context data that is specific for this event.
+   *
+   * @returns - Track event data to be sent to integrations.
+   */
+  override async getTrackEventData(
+    type: TrackTypesValues,
+    event: string,
+    properties?: EventProperties,
+    eventContext?: ContextData,
+  ): Promise<EventData<TrackTypesValues>> {
+    this.processContext(eventContext);
+
+    return await super.getTrackEventData(type, event, properties, eventContext);
+  }
+
+  /**
+   * Getter for the context object.
+   *
+   * @param key - Key to retrieve from the context. If not specified, will return the whole data stored
+   *              in the context.
+   *
+   * @returns Value for the key in context or the whole context data if key is not specified.
+   */
+  override context(): Promise<ContextData>;
+  override context(key: string): Promise<unknown>;
+  override async context(key?: string) {
+    const context = await super.context();
+
+    this.processContext(context);
+
+    return key ? get(context, key) : context;
+  }
+
+  /**
+   * Process context to append context features related with this package.
+   *
+   * @param context - Context data that is specific for this event.
+   */
+  processContext(context?: ContextData) {
+    if (context) {
+      context.library = {
+        name: PACKAGE_NAME,
+        version: `${context.library.name}@${context.library.version};${PACKAGE_NAME}@${PACKAGE_VERSION};`,
+      };
     }
   }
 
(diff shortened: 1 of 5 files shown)


nelsonleite (approved):

packages/analytics/src/Analytics.ts
```diff
@@ -32,11 +35,21 @@ import type {
 } from './types/analytics.types.js';
 import type { Storage } from './utils/types/index.js';
 
+// We use a require here to avoid typescript complaining of `package.json` is not
+// under rootDir that we would get if we used an import. Typescript apparently ignores
+// requires.
+const {
+  name: PACKAGE_NAME,
```
boliveira, line 42: These are already being exported from the `utils/constants.js` module. You can update the import you already have with that.

packages/react/src/analytics/integrations/GA4/GA4.ts
```diff
@@ -399,7 +408,18 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
...
+        data,
+        `context.event.${utils.ANALYTICS_PACKAGE_VERSION}`,
+        '',
+      ) as string;
+
+      const packageVersion = `${PACKAGE_NAME}@${PACKAGE_VERSION};`;
+
       eventProperties[utils.ANALYTICS_UNIQUE_EVENT_ID] = uniqueEventId;
+      eventProperties[
+        'analytics_package_version'
```
boliveira, line 421: Why not add 2 separate properties, one for each version?
Also, the properties should start with a `__` to indicate that they are internal.

packages/react/src/analytics/integrations/GA4/GA4.ts
```diff
@@ -399,7 +408,18 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
         `context.event.${utils.ANALYTICS_UNIQUE_EVENT_ID}`,
       ) as string;
 
+      const coreVersion = get(
+        data,
+        `context.event.${utils.ANALYTICS_PACKAGE_VERSION}`,
+        '',
+      ) as string;
+
+      const packageVersion = `${PACKAGE_NAME}@${PACKAGE_VERSION};`;
```
boliveira, line 417: This can be calculated outside of the track method. In fact, you could add this calculation in the analytics facade for the react and pass it in the event context just the way that the Analytics class in analytics package does. That way you can share that value with all integrations that need it.

packages/analytics/src/Analytics.ts
```diff
@@ -687,6 +701,7 @@ class Analytics {
       event: {
         ...eventContext,
         [ANALYTICS_UNIQUE_EVENT_ID]: uuid(),
+        [ANALYTICS_PACKAGE_VERSION]: this.analyticsVersion,
```
boliveira, line 704: Don't we have a library property in the event data already that contains this info? Why should it be specific to the event?

packages/react/src/analytics/integrations/GA4/GA4.ts
danielbento92: Due last tests, it was informed that properties which begin with _ or __ aren't appear on GA4 Data, so in future we will replace these Analytics_Unique_Event_Id value or update only in GA4 that were this would be an issue.

About 2 separate properties, it's because it was requested to be created a field, named as analytics_package_version to set versions used.

boliveira: Interesting, did not know about those rules for event parameter names but I see they are documented here: https://support.google.com/analytics/answer/13316687#zippy=%2Cweb.

At least then we could use some kind of prefix to make it easier to understand, like `META_analytics_package_version` or something like that.

packages/react/src/analytics/analytics.ts
```diff
@@ -56,6 +66,47 @@ class AnalyticsWeb extends Analytics {
...
+  ): Promise<EventData<TrackTypesValues>> {
+    this.processContext(eventContext);
+
+    return await super.getTrackEventData(type, event, properties, eventContext);
+  }
+
+  override context(): Promise<ContextData>;
+  override context(key: string): Promise<unknown>;
+  override async context(key?: string) {
+    if (key && key !== 'library.version') {
```
boliveira, line 83: Why is this if necessary?

packages/react/src/analytics/analytics.ts
```diff
@@ -56,6 +66,47 @@ class AnalyticsWeb extends Analytics {
...
+    const context = await super.context();
+
+    this.processContext(context);
+
+    // if key are set and goes to this path, then key must be equal to analyticsVersion
+    return key ? get(context, key) : context;
+  }
+
+  processContext(context?: ContextData) {
+    if (context?.library) {
```
boliveira, line 96: Context and library will always be defined, so this check is not necessary.

packages/react/src/analytics/analytics.ts
```diff
@@ -56,6 +66,47 @@ class AnalyticsWeb extends Analytics {
...
+      let analyticsVersions = '';
+
+      if (context?.library?.name && context?.library?.version) {
```
boliveira, line 99: Same here, `library` will always be set in this case so as name and version as well.

packages/react/src/analytics/analytics.ts
danielbento92: The idea here it's to return original function when some key are set, because it would return only the value from context using the key (`return key? get(data, key) : data `).

In this new version, we are intercepting this method on react Analytic's to change library version property.
In this case, we can just assume that if key are set and key are different than prop changed, we could return directly output from super.context, otherwise, we need to retrieve super.context data, and process data to change this prop (library.version)

packages/react/src/analytics/analytics.ts
danielbento92: Following the code, this method are created to apply same code in two different places, avoiding duplication.

One of the calls, are on getTrackEventData, which context parameter could be undefined, so In this method I need to check if context are filled.

But I could remove check of library, because if context are assigned, by contract, it requires library property are set too.

packages/react/src/analytics/analytics.ts
danielbento92: Fixed

packages/react/src/analytics/analytics.ts
danielbento92: I've checked, Analytics file (@farfetch/blackout-analytics), and when it passed key value, only use them on last moment, to retrieve data from that key.

To simplify, I can discard use key on super.context call, doing this get on react analytics file (after process library values).

PR #741, opened 2023-04-27
feat: setup analytics api integration next

talbertosilva:
## Description

Setup of analytics api integration on next

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/analytics/src/utils/getters.ts
+++ b/packages/analytics/src/utils/getters.ts
@@ -110,3 +110,19 @@ export const getLocation = (
     Record<string, string | undefined>
   >;
 };
+
+/**
+ * Obtain cookie value using the cookie name.
+ *
+ * @param name - Cookie name.
+ *
+ * @returns The cookie value.
+ */
+export const getCookie = (name: string): string | void => {
+  const value = `; ${document.cookie}`;
+  const parts = value.split(`; ${name}=`);
+
+  if (parts.length === 2) {
+    return parts.pop()?.split(';').shift();
+  }
+};
--- a/packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
+++ b/packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
@@ -1,0 +1,110 @@
+import {
+  type AnalyticsApiIntegrationOptions,
+  type AnalyticsApiTrackData,
+} from './types/types.js';
+import { client } from '@farfetch/blackout-client';
+import {
+  type EventData,
+  EventTypes,
+  integrations,
+  type LoadIntegrationEventData,
+  PageTypes,
+  type StrippedDownAnalytics,
+  type TrackTypesValues,
+  utils,
+} from '@farfetch/blackout-analytics';
+import { isArray, omit } from 'lodash-es';
+
+export default class AnalyticsAPI extends integrations.Integration<AnalyticsApiIntegrationOptions> {
+  static override [utils.CONSENT_CATEGORIES_PROPERTY] =
+    utils.DefaultConsentKeys.MARKETING;
+  private debugMode: boolean | undefined;
+  private whitelisted: Array<string>;
+
+  /**
+   * Creates an instance of Analytics Api integration.
+   *
+   * @param options               - User configured options.
+   * @param loadData              - Analytics's load event data.
+   * @param strippedDownAnalytics - Analytics stripped down instance.
+   */
+  constructor(
+    options: AnalyticsApiIntegrationOptions,
+    loadData: LoadIntegrationEventData,
+    strippedDownAnalytics: StrippedDownAnalytics,
+  ) {
+    super(options, loadData, strippedDownAnalytics);
+
+    if (
+      this.options.whitelistedEvents &&
+      !isArray(this.options.whitelistedEvents)
+    ) {
+      throw new Error(
+        '[Analytics API] - The value `options.whitelistedEvents` from Analytics Api integration options must be an array.',
+      );
+    }
+
+    if (
+      this.options.blacklistedEvents &&
+      !isArray(this.options.blacklistedEvents)
+    ) {
+      throw new Error(
+        '[Analytics API] - The value `options.blacklistedEvents` from Analytics Api integration options must be an array.',
+      );
+    }
+
+    this.debugMode =
+      utils.getCookie('analyticsAPIDebug')?.toLowerCase() === 'true';
+    this.whitelisted = this.options.whitelistedEvents || [
+      ...Object.values(PageTypes),
+      ...Object.values(EventTypes),
+    ];
+  }
+
+  /**
+   * Track method that will call Analytics API with the necessary data per event.
+   * If there's more than one pageType per event, it will call Analytics API with each page type at a time.
+   *
+   * @param data - Track event data.
+   *
+   * @returns Promise that will resolve when the method finishes.
+   */
+  override async track(data: EventData<TrackTypesValues>) {
+    if (
+      this.whitelisted?.includes(data.event) &&
+      !this.options.blacklistedEvents?.includes(data.event)
+    ) {
+      const payload = {
+        data,
+        options: this.getOptions(),
+      };
+
+      if (this.debugMode) {
+        utils.logger.info('[Analytics API] - track:', payload);
+      }
+
+      return await this.callAPI(payload);
+    }
+  }
+
+  /**
+   * Calls the Analytics API with the given payload.
+   *
+   * @param data - Analytics' event data.
+   */
+  async callAPI(data: AnalyticsApiTrackData) {
+    await client.post('/analytics/v1/events', data);
+  }
+
+  /**
+   * Obtain options with additional common properties.
+   *
+   * @returns The options data.
+   */
+  private getOptions() {
+    return omit({ ...this.options, debugMode: this.debugMode }, [
+      'whitelistedEvents',
+      'blacklistedEvents',
+    ]);
+  }
+}
--- a/packages/react/src/analytics/integrations/AnalyticsApi/__tests__/AnalyticsApi.test.ts
+++ b/packages/react/src/analytics/integrations/AnalyticsApi/__tests__/AnalyticsApi.test.ts
@@ -1,0 +1,189 @@
+import { AnalyticsApi } from '../../index.js';
+import { client } from '@farfetch/blackout-client';
+import {
+  EventTypes,
+  integrations,
+  type LoadIntegrationEventData,
+  type PageTypes,
+  type StrippedDownAnalytics,
+  utils,
+} from '@farfetch/blackout-analytics';
+import {
+  loadIntegrationData,
+  trackEventsData,
+} from 'tests/__fixtures__/analytics/index.mjs';
+import type { AnalyticsApiIntegrationOptions } from '../types/types.js';
+
+utils.logger.error = jest.fn();
+
+utils.logger.warn = jest.fn();
+
+Object.defineProperty(document, 'cookie', {
+  writable: true,
+  value: 'analyticsAPIDebug=true',
+});
+
+jest.mock('@farfetch/blackout-client', () => ({
+  ...jest.requireActual('@farfetch/blackout-client'),
+  client: { post: jest.fn() },
+}));
+
+const strippedDownAnalytics: StrippedDownAnalytics = {
+  createEvent: type => Promise.resolve({ ...loadIntegrationData, type }),
+};
+
+const defaultTrackEventData = trackEventsData[EventTypes.PRODUCT_ADDED_TO_CART];
+
+function createAnalyticsApiInstance(
+  options: AnalyticsApiIntegrationOptions,
+  loadData: LoadIntegrationEventData = loadIntegrationData,
+  analytics: StrippedDownAnalytics = strippedDownAnalytics,
+) {
+  return AnalyticsApi.createInstance(
+    options,
+    loadData,
+    analytics,
+  ) as AnalyticsApi;
+}
+
+async function createAnalyticsApiInstanceAndLoad(
+  options: AnalyticsApiIntegrationOptions,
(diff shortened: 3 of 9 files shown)


danielbento92 (approved):

nelsonleite (approved):

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
```diff
@@ -0,0 +1,97 @@
...
+    super(options, loadData, strippedDownAnalytics);
+
+    // Check if options don't exists
+    if (!options) {
+      throw new Error(
+        '[Analytics API] - Analytics Api integration options must be set.',
+      );
+    }
+
+    if (!isArray(options.conversionsAPI?.eventsList)) {
```
boliveira, line 42:
```suggestion
    if (!isArray(options.conversionsAPI.eventsList)) {
```

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
```diff
@@ -0,0 +1,97 @@
...
+    await client.post('/analytics/v1/events', data);
+  }
+
+  /**
+   * Obtain options with additional common properties.
+   *
+   * @returns The options data.
+   */
+  private getOptions() {
+    return { ...this.options, debugMode: this.debugMode };
```
boliveira, line 95: Question: Why is it necessary to send the `debugMode` property to the server as it is derived from a cookie? Wouldn't the server receive the cookie as well and derive the `debugMode` from it?

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
```diff
@@ -0,0 +1,97 @@
...
+   * Track method that will call Analytics API with the necessary data per event.
+   * If there's more than one pageType per event, it will call Analytics API with each page type at a time.
+   *
+   * @param data - Track event data.
+   *
+   * @returns Promise that will resolve when the method finishes.
+   */
+  override async track(data: EventData<TrackTypesValues>) {
+    if (this.debugMode) {
+      utils.logger.info('ANALYTICS API track:', {
```
boliveira, line 62: You could create the `data` payload just once instead of creating it here and in the call to `this.callAPI` like this:

```js
const isEventInWatchEventList = (this.options.conversionsAPI.eventsList as Array<string>).includes(data.event);

if(isEventInWatchEventList || this.debugMode) {
  const payload = {
        data,
        options: this.getOptions(),
   });
  
  if(this.debugMode) {
    utils.logger.info('ANALYTICS API track:', payload);
  }

  if(isEventInWatchEventList) {
   return await this.callAPI(payload);
  }

}


```

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  /**
+   * Calls the Analytics API with the given payload.
+   *
+   * @param data - Analytics' event data.
+   */
+  async callAPI(data: AnalyticsApiTrackData) {
+    await client.post('/analytics/v1/events', data);
```
boliveira, line 86: Shouldn't we create a client in `@farfetch/blackout-client` and use it as we do with Omnitracking? Also, it would be interesting to allow the user to specify the client he wants and only use the default if he does not specify anything.

packages/react/src/analytics/integrations/AnalyticsApi/__tests__/AnalyticsApi.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    const loadData: LoadIntegrationEventData = {
+      ...loadIntegrationData,
+      user: {
+        ...loadIntegrationData.user,
+        id: 12345678,
+      },
+    };
+
+    it('Should return a AnalyticsApi instance from createInstance', () => {
```
boliveira, line 100:
```suggestion
    it('Should return an AnalyticsApi instance from createInstance', () => {
```

packages/react/src/analytics/integrations/AnalyticsApi/__tests__/AnalyticsApi.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      it('Should map the mocked event', async () => {
+        analyticsApiInstance = await createAnalyticsApiInstanceAndLoad(
+          validOptions,
+        );
+
+        await analyticsApiInstance.track(defaultTrackEventData);
+
+        expect(client.post).toHaveBeenCalledWith(
+          '/analytics/v1/events',
+          expect.objectContaining({
```
boliveira, line 156: Why not check for the options property as well?

packages/react/src/analytics/integrations/AnalyticsApi/types/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  type IntegrationOptions,
+  type LoadIntegrationEventData,
+  type PageTypes,
+  type SetUserEventData,
+  type TrackTypesValues,
+} from '@farfetch/blackout-analytics';
+
+export interface AnalyticsApiIntegrationOptions extends IntegrationOptions {
+  conversionsAPI: {
+    eventsList: Array<`${EventTypes}` | `${PageTypes}`>;
```
boliveira, line 13:
```suggestion
    eventList: Array<string>; // The user could define new event/pages types that are not in `EventTypes` or `PageTypes`
```

packages/react/src/analytics/integrations/AnalyticsApi/types/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    testEventCode?: string;
+  };
+}
+
+export type AnalyticsApiTrackData = {
+  data:
```
boliveira, line 19: From the implementation code, I believe only `EventData<TrackTypesValues>` will be possible to be sent. Should this integration handle `onSetUser` and `loadIntegration` events as well? If so, then the implementation must be changed to handle that case (track method will not be called by analytics in those cases).

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
danielbento92: Debug mode, will be used for two things: first it's for log events on frontend layer.
The second it's related with AnalyticsApi server, witch are prepared to receive events, then return 200 (performance). Debug Mode flag, it's a flag added to await for internal responses and return more data. 

The point here isn't to set debug property on server side, and passed by frontend, that allow an easy way of setting debug only for developers or testers when they need that.

packages/react/src/analytics/integrations/AnalyticsApi/types/types.ts
danielbento92: For this point, we have discussed this already and at the moment server side don't require these events. Of course, we could implement as well but it will be increased requests that won't be processed by server.

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
boliveira: I am not sure if I understood your comment. Debug mode is being set by the integration if the cookie `analyticsAPIDebug` is set. If the cookie is set, then the server will receive it when the POST request is made so it is redundant to send the `debugMode` in the payload of the POST request.

packages/react/src/analytics/integrations/AnalyticsApi/types/types.ts
danielbento92: ConversionsAPI only deal with some specific events, so I think it's pointless open this for any string considering eventTypes and pageTypes has all possibilities listed.

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
talbertosilva: This debugMode parameter is received from a cookie but needs to be included in the request body since the server side is ready to answer to this parameter through the body

boliveira: So the body of the request has to contain the `debugMode` parameter inside the `options` parameter or else the server will not assume it should treat the request in `debugMode` even though it receives the same cookie with it? Is that correct?

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
danielbento92: I'm not sure if this case are already cover, but this initiative has some improvements in future, due delivery time. I think we could postpone this as well, and create an task for latter.

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
danielbento92: In the AnalyticsAPI, we receive a single parameter, with data (the event data itself) and options inside. DebugMode is one of this properties that would be send in data.

packages/analytics/src/utils/getters.ts
```diff
@@ -110,3 +110,21 @@ export const getLocation = (
...
+ */
+export const getCookie = (name: string) => {
+  const value = `; ${document.cookie}`;
+  const parts = value.split(`; ${name}=`);
+
+  if (parts.length === 2) {
+    return parts.pop()?.split(';').shift();
+  }
+
+  return;
```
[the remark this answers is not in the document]
talbertosilva: Fixed!

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
talbertosilva: Fixed!

packages/react/src/analytics/integrations/AnalyticsApi/__tests__/AnalyticsApi.test.ts
talbertosilva: Fixed!

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Track method that will call Analytics API with the necessary data per event.
+   * If there's more than one pageType per event, it will call Analytics API with each page type at a time.
+   *
+   * @param data - Track event data.
+   *
+   * @returns Promise that will resolve when the method finishes.
+   */
+  override async track(data: EventData<TrackTypesValues>) {
+    const whitelisted = this.options.whitelistedEvents || [
```
boliveira, line 68: Instead of doing this every time the track method is called you can set this in a private variable of the instance in the constructor after you validated the options argument.

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
```diff
@@ -0,0 +1,107 @@
...
+  type EventData,
+  EventTypes,
+  integrations,
+  type LoadIntegrationEventData,
+  PageTypes,
+  type StrippedDownAnalytics,
+  type TrackTypesValues,
+  utils,
+} from '@farfetch/blackout-analytics';
+import { isArray } from 'lodash';
```
boliveira, line 16: Do not use lodash, you need to use `lodash-es` always since this package is ESM-only now.

```suggestion
import { isArray } from 'lodash-es';
```

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
boliveira: After the changes we discussed yesterday you should not send the options passed by the user to the server as it is unnecessary. Only pass the debugMode property.

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
```diff
@@ -0,0 +1,107 @@
...
+    ) as Array<string>;
+
+    if (filteredEvents.includes(data.event)) {
+      const payload = {
+        data,
+        options: this.getOptions(),
+      };
+
+      if (this.debugMode) {
+        utils.logger.info('ANALYTICS API track:', payload);
```
boliveira, line 83: Use the same format you use for other messages:

```suggestion
        utils.logger.info('[Analytics API] - track:', payload);
```

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
danielbento92: I don't remember of discarding options, and if it's to remove options, it will be need to change service that is already in live. I think we should maintain options, avoiding create breaking changes.

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @param data - Track event data.
+   *
+   * @returns Promise that will resolve when the method finishes.
+   */
+  override async track(data: EventData<TrackTypesValues>) {
+    const whitelisted = this.options.whitelistedEvents || [
+      ...Object.values(PageTypes),
+      ...Object.values(EventTypes),
+    ];
+    const filteredEvents = whitelisted?.filter(
```
boliveira, line 72: Let's simplify this, you only need to check if the passed event is in the `whitelistedEvents` and not in the `blacklistedEvents`, this code is unnecessary. You only need this:

```js
if (whitelistedEvents.includes(data.event) && !blacklistedEvents.includes(data.event)) {
   // Send event to server...
```

boliveira (approved):

PR #742, opened 2023-04-27
feat: setup analytics api integration

talbertosilva:
## Description

Setup of analytics api integration on next

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.js
+++ b/packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.js
@@ -1,0 +1,113 @@
+import { integrations, utils } from '@farfetch/blackout-core/analytics';
+import client from '@farfetch/blackout-core/helpers/client';
+import eventTypes from '../../../../../core/src/analytics/types/eventTypes';
+import isArray from 'lodash/isArray';
+import omit from 'lodash/omit';
+import pageTypes from '../../../../../core/src/analytics/types/pageTypes';
+
+export default class AnalyticsAPI extends integrations.Integration {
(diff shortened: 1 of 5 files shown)


packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.js
```diff
@@ -0,0 +1,101 @@
...
+    if (
+      this.options.conversionsAPI.eventsList &&
+      this.options.conversionsAPI.eventsList.includes(data.event)
+    ) {
+      return await this.callAPI({
+        data,
+        options: this.getOptions(),
+      });
+    }
+    if (this.debugMode) {
```
ruifcnunes, line 53: this condition should be inside the previous if, otherwise all tracks that will not reach the analytics API will also be logged.

packages/react/src/analytics/integrations/AnalyticsApi/AnalyticsApi.js
```diff
@@ -0,0 +1,101 @@
...
+    this.debugMode = utils.getCookie('analyticsAPIDebug');
+  }
+
+  /**
+   * Creates and returns the instance for the Analytics API integration.
+   *
+   * @param {object} options - Integration options.
+   * @param {object} loadData - Analytics' load event data.
+   * @param {object} analytics - Analytics stripped down instance.
+   * @param strippedAnalytics
```
ruifcnunes, line 26: Extra @param can be removed. Also, ensure there's an empty line between the last `@param` and `@returns` and that the last argument is document with the name `strippedAnalytics` instead of `analytics`

danielbento92 (approved):

PR #756, opened 2023-05-10
Feat analytics improvements

danielbento92:
## Description

- add locationId, parameter on bag and wishlist middlewares;
- add total parameter on some GTM events.

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/analytics/integrations/GTM/eventsMapper.ts
+++ b/packages/react/src/analytics/integrations/GTM/eventsMapper.ts
@@ -88,6 +88,8 @@ const eventsMapper: EventMappers = {
     return {
       products: get(properties, 'products', []).map(getProductData),
       orderId: properties.orderId,
+      value: 'total' in properties ? properties.total : properties.value,
+      total: properties.total,
     };
   },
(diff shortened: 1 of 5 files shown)


talbertosilva (approved):

packages/react/src/analytics/integrations/GTM/eventsMapper.ts
```diff
@@ -88,6 +88,8 @@ const eventsMapper: EventMappers = {
     return {
       products: get(properties, 'products', []).map(getProductData),
       orderId: properties.orderId,
+      total: properties.total,
```
boliveira, line 91: What is supposed to be the difference between `total` and `value` in checkout events ?

danielbento92: I look into other integrations, and I saw that we are using total instead of value, but in some cases, we already have value mapped, such in EventTypes.PLACE_ORDER_STARTED.

To avoid breaking changes, and by according with our documentation, the value of grandTotal in checkout, would be fill the "Total" in analytics contract. So in this case, I maintain this two options, only to maintain compatibility.

boliveira (approved):

PR #766, opened 2023-05-18
feat(client): add backorder types to order item

andreecosta:
## Description

- This adds `expectedFulfillmentDate` and `saleIntent` types to order item.

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/client/src/orders/types/orderItem.types.ts
+++ b/packages/client/src/orders/types/orderItem.types.ts
@@ -14,9 +14,7 @@ import type {
   ProductVariantAttribute,
 } from '../../index.js';
 
-import { SaleIntent as SelectedSaleIntent } from '../../types/common/index.js';
-
-export { SelectedSaleIntent };
+import { type SaleIntent } from '../../types/common/index.js';
 
 export type ShippingService = {
   description: string;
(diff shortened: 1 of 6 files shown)


packages/client/src/orders/types/orderItem.types.ts
```diff
@@ -87,6 +87,8 @@ export type OrderItem = {
   productType: keyof typeof ProductType;
   selectedSaleIntent?: SelectedSaleIntent | string;
   metadata?: Record<string, string>;
+  expectedFulfillmentDate?: string;
+  saleIntent?: string;
```
boliveira, line 91: What is the difference between `saleIntent` and `selectedSaleIntent` that is also defined in this `OrderItem` type?

nelsonleite (approved):

packages/client/src/orders/types/orderItem.types.ts
andreecosta: As we talked in private, I've checked with the BE team and the `selectedSaleIntent` property does not exist. So it was removed to only keep `saleIntent`. I've also renamed the type since having the `selected` keyword no longer makes sense. Please validate if everything is ok 🙂

boliveira (approved):

PR #789, opened 2023-06-12
feat: implement eircode validation

tiagodicosta:
## Description

Create a function and utils to validate if a zip code from a shipping address is valid.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/checkout/hooks/useCheckout.ts
+++ b/packages/react/src/checkout/hooks/useCheckout.ts
@@ -17,8 +17,10 @@ import {
   updateCheckoutOrder,
 } from '@farfetch/blackout-redux';
 import {
+  type CheckoutAddress,
   type CheckoutOrder,
   type Config,
+  type CountryAddressSchema,
   type GetCheckoutOrderQuery,
   type GetCollectPointsQuery,
   HttpHeaders,
@@ -63,7 +65,6 @@ function useCheckout(
     createData,
     createConfig,
     enableAutoCreate = false,
-    isoCodesToValidate = [],
   } = options;
 
   const isCheckoutOrderLoading = useSelector(isCheckoutOrderLoadingSelector);
@@ -489,30 +490,33 @@ function useCheckout(
     { enableAutoFetch: false },
   );
 
-  const isShippingAddressZipCodeValid = useCallback(async () => {
-    let addressSchema = countryAddressSchemas;
+  const isShippingAddressZipCodeValid = useCallback(
+    async (
+      shippingAddress:
+        | CheckoutAddress
+        | undefined = checkoutOrder?.shippingAddress,
+    ) => {
+      let addressSchema: CountryAddressSchema[] | undefined =
+        countryAddressSchemas;
 
-    if (!checkoutOrder?.shippingAddress) {
-      return { isValid: false };
-    }
+      if (!shippingAddress) {
+        return { isValid: false };
+      }
 
-    if (!addressSchema) {
-      addressSchema = await fetchCountryAddressSchemas(
-        checkoutOrder?.shippingAddress?.country.alpha2Code,
-      );
-    }
+      if (!addressSchema) {
+        addressSchema = await fetchCountryAddressSchemas(
+          shippingAddress?.country.alpha2Code,
+        );
+      }
 
-    return validateShippingAddressZipCode(
+      return validateShippingAddressZipCode(shippingAddress, addressSchema);
+    },
+    [
+      countryAddressSchemas,
       checkoutOrder?.shippingAddress,
-      addressSchema,
-      isoCodesToValidate,
-    );
-  }, [
-    checkoutOrder?.shippingAddress,
-    countryAddressSchemas,
-    fetchCountryAddressSchemas,
-    isoCodesToValidate,
-  ]);
+      fetchCountryAddressSchemas,
+    ],
+  );
 
   useEffect(() => {
     if (
--- a/packages/react/src/checkout/hooks/useCheckout.ts
+++ b/packages/react/src/checkout/hooks/useCheckout.ts
@@ -33,6 +33,10 @@ import {
   type PutPaymentIntentInstrumentData,
 } from '@farfetch/blackout-client';
 import { useCallback, useEffect, useMemo } from 'react';
+import {
+  useCountryAddressSchemas,
+  validateShippingAddressZipCode,
+} from '../../index.js';
 import { useSelector } from 'react-redux';
 import { useUser } from '../../users/index.js';
 import useAction from '../../helpers/useAction.js';
@@ -59,6 +63,7 @@ function useCheckout(
     createData,
     createConfig,
     enableAutoCreate = false,
+    isoCodesToValidate = [],
   } = options;
 
   const isCheckoutOrderLoading = useSelector(isCheckoutOrderLoadingSelector);
@@ -476,6 +481,39 @@ function useCheckout(
     reset();
   }
 
+  const {
+    data: { countryAddressSchemas },
+    actions: { fetch: fetchCountryAddressSchemas },
+  } = useCountryAddressSchemas(
+    (checkoutOrder?.shippingAddress?.country.alpha2Code as string) || '',
+    { enableAutoFetch: false },
+  );
+
+  const isShippingAddressZipCodeValid = useCallback(async () => {
+    let addressSchema = countryAddressSchemas;
+
+    if (!checkoutOrder?.shippingAddress) {
+      return { isValid: false };
+    }
+
+    if (!addressSchema) {
+      addressSchema = await fetchCountryAddressSchemas(
+        checkoutOrder?.shippingAddress?.country.alpha2Code,
+      );
+    }
+
+    return validateShippingAddressZipCode(
+      checkoutOrder?.shippingAddress,
+      addressSchema,
+      isoCodesToValidate,
+    );
+  }, [
+    checkoutOrder?.shippingAddress,
+    countryAddressSchemas,
+    fetchCountryAddressSchemas,
+    isoCodesToValidate,
+  ]);
+
   useEffect(() => {
     if (
       enableAutoCreate &&
@@ -547,6 +585,7 @@ function useCheckout(
       isOrderConfirmed,
       isOrderAwaitingPayment,
       getSelectedShippingOption,
+      isShippingAddressZipCodeValid,
     },
     data,
     isAnythingLoading,
--- a/packages/react/src/helpers/__tests__/validateShippingAddressZipCode.test.ts
+++ b/packages/react/src/helpers/__tests__/validateShippingAddressZipCode.test.ts
@@ -1,0 +1,134 @@
+import { type AddressType } from '@farfetch/blackout-client';
+import { validateShippingAddressZipCode } from '../index.js';
+
+const eircodeRegex =
+  '^(?=.{7,8}$)^([AC-FHKNPRTV-Y]{1}[0-9]{2}|D6W)[ ]?[0-9AC-FHKNPRTV-Y]{4}$';
+
+describe('validateShippingAddressZipCode', () => {
+  const mockAddressSchema = [
+    {
+      addressSchemaLines: [
+        {
+          id: '9a24c613-18c5-4b0d-a789-d79369777262',
+          parentId: '00000000-0000-0000-0000-000000000000',
+          name: 'Postal Code',
+          position: 0,
+          type: 'FreeText',
+          validationRegex: eircodeRegex,
+          apiMapping: 'ZipCode',
+          isMandatory: true,
+          maxLength: 45,
+          minLength: 1,
+          column: 0,
+          row: 0,
+        },
+      ],
+      addressType: 'Shipping' as AddressType,
+    },
+  ];
+
+  it("should return true if the country isn't one of the countries to validate", async () => {
+    const mockAddress = {
+      addressLine1: 'Portugal Address',
+      city: {
+        countryId: 0,
+        id: 0,
+        name: 'Portugal',
+      },
+      country: {
+        alpha2Code: 'PT',
+        id: 98,
+        name: 'Portugal',
+        continentId: 1,
+      },
+      zipCode: '4444',
+    };
+
+    expect(
+      await validateShippingAddressZipCode(mockAddress, mockAddressSchema, [
+        'ES',
+      ]),
+    ).toEqual({
+      isValid: false,
+      error: 'ER01',
+    });
+  });
+
+  it('should return true if the country is one of the countries to validate and zipCode is valid', async () => {
+    const mockAddress = {
+      addressLine1: 'Ireland Address',
+      city: {
+        countryId: 0,
+        id: 0,
+        name: 'Ireland',
+      },
+      country: {
+        alpha2Code: 'IE',
+        id: 98,
+        name: 'Ireland',
+        continentId: 1,
+      },
+      zipCode: 'D02 N725',
+    };
+
+    expect(
+      await validateShippingAddressZipCode(mockAddress, mockAddressSchema, [
+        'IE',
+      ]),
+    ).toEqual({
+      isValid: true,
+    });
+  });
+
+  it('should return false if the country is one of the countries to validate and zipCode is invalid', async () => {
+    const mockAddress = {
+      addressLine1: 'Ireland Address',
+      city: {
+        countryId: 0,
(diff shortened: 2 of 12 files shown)


packages/client/src/addresses/getAddressSchema.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Method responsible for getting the predictions.
+ *
+ * @param countryCode   - country code to get the schema.
+ * @param config - Custom configurations to send to the client instance (axios).
+ *
+ * @returns Promise that will resolve when the call to the endpoint finishes.
+ */
+const getAddressSchema: GetAddressSchema = (countryCode, config) =>
```
boliveira, line 14: This client is already implemented as `getCountryAddressSchemas`. You can remove it.

packages/react/src/helpers/validateEircode.ts
```diff
@@ -0,0 +1,35 @@
+import {
+  type AddressBase,
+  type AddressSchema,
+  getAddressSchema,
+} from '@farfetch/blackout-client';
+
+/**
+ * Validates eircode and returns true or false depending on the test result..
```
boliveira, line 8:
```suggestion
 * Validates eircode and returns true if the passed address' zipcode is an eircode, false otherwise.
```

packages/react/src/helpers/validateEircode.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @param address - address that will be tested the eircode.
+ *
+ * @returns - boolean which is the result of the validation.
+ * */
+export default async function validateEircode(
+  address: AddressBase,
+): Promise<boolean> {
+  const countryCode = address.country.alpha2Code;
+  const addressSchema: AddressSchema = await getAddressSchema(countryCode);
```
boliveira, line 18: I think it is better to remove the call to the client here and accept a parameter with the country address schemas. This will change the return value of the function as well to be a `boolean` instead of a `Promise<boolean>`.

tests/__fixtures__/addresses/addresses.fixtures.mts
```diff
@@ -1,10 +1,151 @@
...
+  AddressSchemaLineType,
+  AddressType,
+} from '@farfetch/blackout-client';
 
 export const addressId = '1234';
 export const addressId2 = '2222222';
 export const addressId3 = '33333333';
 export const userId = 123456;
 
+export const mockAddressSchema: AddressSchema = [
```
boliveira, line 13: There is a mock already inside `locale.fixtures.mts` (`mockGetAddressSchemaResponse`). Reuse it,please.

packages/react/src/helpers/__tests__/validateEircode.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  getAddressSchema: jest.fn(),
+}));
+
+const mockGetAddressSchema = jest.mocked(getAddressSchema);
+const eircodeRegex =
+  '^(?=.{7,8}$)^([AC-FHKNPRTV-Y]{1}[0-9]{2}|D6W)[ ]?[0-9AC-FHKNPRTV-Y]{4}$';
+
+describe('validateEirocde', () => {
+  beforeEach(() => {
+    mockGetAddressSchema.mockResolvedValueOnce([
```
nelsonleite, line 18: I believe you don't need to do this on `beforeEach`, you can set the value right on the mock (line 9) or at least have the `mockResolvedValue` outside 🤔

packages/react/src/checkout/hooks/useCheckout.ts
```diff
@@ -476,6 +481,39 @@ function useCheckout(
...
+  );
+
+  const isShippingAddressZipCodeValid = useCallback(async () => {
+    let addressSchema = countryAddressSchemas;
+
+    if (!checkoutOrder?.shippingAddress) {
+      return { isValid: false };
+    }
+
+    if (!addressSchema) {
```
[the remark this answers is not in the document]
tiagodicosta: Not really, because I'm sending `{ enableAutoFetch: false }`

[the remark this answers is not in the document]
nelsonleite: And with auto fetch being false, is it fair to fetch automatically fetch if there is no address schema? 🤔 Is it the expected behaviour?

packages/react/src/helpers/validateShippingAddressZipCode.ts
```diff
@@ -0,0 +1,58 @@
...
+  address: AddressBase,
+  addressSchema: CountryAddressSchema[],
+  isoCodesToValidate: string[],
+): { isValid: boolean; error?: string } {
+  if (!(isoCodesToValidate || []).includes(address.country.alpha2Code)) {
+    return { isValid: true };
+  }
+
+  const schema =
+    addressSchema.find(({ addressType }) => addressType === 'Shipping') ||
```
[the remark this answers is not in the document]
ivomota: BTW, I prefer the 2nd one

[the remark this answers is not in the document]
tiagodicosta: I dont think so. I want to check first if the `shipping` type exists, if not I will check if `Any` exists.

packages/react/src/checkout/hooks/useCheckout.ts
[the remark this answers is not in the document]
tiagodicosta: Yes, because if there is no address schema yet, I will fetch that, doesn't it make sense?

packages/react/src/helpers/validateShippingAddressZipCode.ts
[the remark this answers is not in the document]
ivomota: That's exactly what this code does. You can change the order of the data in the array ['Shipping', 'Any']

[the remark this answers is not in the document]
tiagodicosta: Understood, I already change it, pls double check

packages/react/src/checkout/hooks/useCheckout.ts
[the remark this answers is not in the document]
nelsonleite: All right, I'm grasping the logic now ✅

ivomota (approved):

packages/react/src/helpers/validateShippingAddressZipCode.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @param address - address
+ * @param addressSchema - address schema to test.
+ * @param isoCodesToValidate - iso codes to be validated.
+ *
+ * @returns - boolean which is the result of the validation.
+ * */
+export default function validateShippingAddressZipCode(
+  address: AddressBase,
+  addressSchema: CountryAddressSchema[],
```
boliveira, line 30: Since this is an array, this should be called addressSchemas. Please, change all references to the plural (don't forget the jsdoc as well)

packages/react/src/helpers/validateShippingAddressZipCode.ts
```diff
@@ -0,0 +1,58 @@
...
+} from '@farfetch/blackout-client';
+
+/**
+ * Zip Code Validation Errors
+ * * ER01 - The zip code validation is ignored because is not included in the countries array to be validated
+ * * ER02 - Zip code schema not found
+ * * ER03 - The zip code inserted is invalid
+ */
+export enum ZipCodeValidationError {
+  ZipCodeValidationCountryIgnored = 'ER01',
```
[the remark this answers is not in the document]
boliveira: This error should be used in the first check (if the isoCodesToValidate contains the shipping address) and it should return false and not true in that case and the error property should be this one.

packages/react/src/helpers/validateShippingAddressZipCode.ts
```diff
@@ -0,0 +1,59 @@
...
+  address: AddressBase,
+  addressSchema: CountryAddressSchema[],
+  isoCodesToValidate: string[] = [],
+): { isValid: boolean; error?: string } {
+  if (!isoCodesToValidate.includes(address.country.alpha2Code)) {
+    return { isValid: true };
+  }
+
+  const schema = addressSchema.find(({ addressType }) =>
+    ['Shipping', 'Any'].includes(addressType),
```
boliveira, line 38: Actually I think this is not the same semantics as it was before. If the first addressType is `Any` and the second is `Shipping`, this will return you the `Any` address schema and not the `Shipping` one which should be given priority.

packages/react/src/helpers/index.ts
```diff
@@ -4,11 +4,13 @@ import validatePassword, {
   PasswordValidationError,
   type PasswordValidationResult,
 } from './validatePassword.js';
+import validateShippingAddressZipCode from './validateShippingAddressZipCode.js';
```
boliveira, line 7: You will need to reexport the ZipCodeValidationError as the consumers will need it to understand the type of error.

tiagodicosta: pls double check

packages/react/src/helpers/validateShippingAddressZipCode.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @returns - boolean which is the result of the validation.
+ * */
+export default function validateShippingAddressZipCode(
+  address: AddressBase,
+  addressSchema: CountryAddressSchema[],
+  isoCodesToValidate: string[] = [],
+): { isValid: boolean; error?: string } {
+  if (!isoCodesToValidate.includes(address.country.alpha2Code)) {
+    return { isValid: true };
```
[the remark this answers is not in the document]
tiagodicosta: right, pls double check

packages/react/src/helpers/validateShippingAddressZipCode.ts
tiagodicosta: done, pls double check

packages/client/src/addresses/getAddressSchema.ts
tiagodicosta: done

nelsonleite (approved):

packages/react/src/helpers/validateShippingAddressZipCode.ts
```diff
@@ -0,0 +1,63 @@
...
+ * * ER03 - The zip code inserted is invalid
+ */
+export enum ZipCodeValidationError {
+  ZipCodeValidationCountryIgnored = 'ER01',
+  ZipCodeSchemaNotFoundError = 'ER02',
+  ZipCodeInvalid = 'ER03',
+}
+
+/**
+ * Validates zip code for an array of iso-codes and returns true if the address' country code is not in the array isoCodesToValidate,
```
boliveira, line 20: Adjust the jsdoc please as it does not match the implementation. Also add a note that if the country is not in the isoCodesToValidate, it will return false and the error code `ZipCodeValidationCountryIgnored` so that consumers can distinguish this case.

packages/react/src/helpers/validateShippingAddressZipCode.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @param addressSchemas - address schemas to test.
+ * @param isoCodesToValidate - iso codes to be validated.
+ *
+ * @returns - boolean which is the result of the validation.
+ * */
+export default function validateShippingAddressZipCode(
+  address: AddressBase,
+  addressSchemas: CountryAddressSchema[],
+  isoCodesToValidate: string[] = [],
+): { isValid: boolean; error?: string } {
```
boliveira, line 33:
```suggestion
): { isValid: boolean; error?: ZipCodeValidationError } {
```

packages/react/src/helpers/validateShippingAddressZipCode.ts
boliveira: I think this should be reverted, you are right that the schema for the `zipCode` might not change between those but I think that as a rule, the `Shipping` schema should be given priority always.

tiagodicosta: OK, fine for me. Pls double check

packages/react/src/helpers/validateShippingAddressZipCode.ts
tiagodicosta: pls double check

packages/react/src/helpers/validateShippingAddressZipCode.ts
tiagodicosta: done

packages/react/src/helpers/__tests__/validateShippingAddressZipCode.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      zipCode: '4444',
+    };
+
+    expect(
+      await validateShippingAddressZipCode(mockAddress, mockAddressSchema, [
+        'ES',
+      ]),
+    ).toEqual({
+      isValid: false,
+      error: 'ER01',
```
boliveira, line 53:
```suggestion
      error: ZipCodeValidationError.CountryIgnored,
```

packages/react/src/helpers/validateShippingAddressZipCode.ts
```diff
@@ -0,0 +1,67 @@
...
+} from '@farfetch/blackout-client';
+
+/**
+ * Zip Code Validation Errors
+ * * ER01 - The zip code validation is ignored because is not included in the countries array to be validated
+ * * ER02 - Zip code schema not found
+ * * ER03 - The zip code inserted is invalid
+ */
+export enum ZipCodeValidationError {
+  ZipCodeValidationCountryIgnored = 'ER01',
```
boliveira, line 14:
```suggestion
  CountryIgnored = 'ER01',
```

packages/react/src/checkout/hooks/useCheckout.ts
```diff
@@ -476,6 +481,46 @@ function useCheckout(
...
+
+      if (!tempShippingAddress) {
+        return { isValid: false };
+      }
+
+      if (!addressSchema) {
+        addressSchema = await fetchCountryAddressSchemas(
+          tempShippingAddress?.country.alpha2Code,
+        );
+        console.log(
```
boliveira, line 509: Search for other console.log/traces you might have and remove them, please.

packages/react/src/checkout/hooks/useCheckout.ts
```diff
@@ -476,6 +481,46 @@ function useCheckout(
...
+    data: { countryAddressSchemas },
+    actions: { fetch: fetchCountryAddressSchemas },
+  } = useCountryAddressSchemas(
+    (checkoutOrder?.shippingAddress?.country.alpha2Code as string) || '',
+    { enableAutoFetch: false },
+  );
+
+  const isShippingAddressZipCodeValid = useCallback(
+    async (shippingAddress?: CheckoutAddress) => {
+      let addressSchema;
```
boliveira, line 494:
```suggestion
      let addressSchema = countryAddressSchemas;
```

packages/react/src/helpers/__tests__/validateShippingAddressZipCode.test.ts
tiagodicosta: this error was removed, because the logic has changed and no longer makes sense.

packages/react/src/checkout/hooks/useCheckout.ts
```diff
@@ -476,6 +481,42 @@ function useCheckout(
...
+
+      addressSchema = countryAddressSchemas;
+
+      const tempShippingAddress =
```
boliveira, line 498: You can remove this line and add use the default value for the shippingAddress parameter like this:

```js
async (shippingAddress: CheckoutAddress = checkoutOrder?.shippingAddress) => {
```

packages/react/src/helpers/validateShippingAddressZipCode.ts
tiagodicosta: error message removed.

packages/react/src/checkout/hooks/useCheckout.ts
tiagodicosta: sorry, already removed

packages/react/src/checkout/hooks/useCheckout.ts
tiagodicosta: updated, pls double check

packages/react/src/checkout/hooks/useCheckout.ts
tiagodicosta: done , pls double check

PR #795, opened 2023-06-15
feat(react): add exchanges hooks

andreecosta:
## Description

This adds 3 exchanges hooks, useExchangeFilters, useExchange and useExchangeBookRequest.

### Dependencies

None.

## Checklist

## Disclaimer


packages/react/src/exchanges/hooks/types/useExchange.ts
```diff
@@ -0,0 +1,14 @@
...
+  Config,
+  ExchangeBookRequest,
+  PostExchangeData,
+  PostExchangeFilterData,
+} from '@farfetch/blackout-client';
+
+export type UseExchangeOptions = {
+  enableAutoFetch?: boolean;
+  fetchConfig?: Config;
+  exchangeData?: PostExchangeData;
```
boliveira, line 11: Both `exchangeData` and `exchangeFilterData` should not be passed as options as they are only used in the `create` actions but the user can pass it directly. It would be useful to add those parameters if the hook would do something else with them.

packages/react/src/exchanges/hooks/types/useExchangeBookRequest.ts
```diff
@@ -0,0 +1,10 @@
+import type {
+  Config,
+  PostExchangeBookRequestData,
+} from '@farfetch/blackout-client';
+
+export type UseExchangeBookRequestOptions = {
+  enableAutoFetch?: boolean;
+  fetchConfig?: Config;
+  bookRequestData?: PostExchangeBookRequestData;
```
boliveira, line 9: Same logic as in the useExchange hook, the user needs to specify it in the `create` action and not in the hook itself.

packages/react/src/exchanges/hooks/useExchange.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+    [exchangeData, fetchConfig, createExchange],
+  );
+
+  const {
+    actions: {
+      fetch: fetchExchangeBookRequest,
+      create: createExchangeBookRequest,
+    },
+  } = useExchangeBookRequest(exchangeIdHookParameter, exchangeBookRequestId, {
```
boliveira, line 79:
```suggestion
  } = useExchangeBookRequest(exchangeIdHookParameter, undefined, {
```

packages/react/src/exchanges/hooks/useExchangeFilter.ts
```diff
@@ -0,0 +1,89 @@
...
+      return createExchangeFilter(exchangeFilterData, config);
+    },
+    [exchangeFilterDataHookParameter, fetchConfig, createExchangeFilter],
+  );
+
+  useEffect(() => {
+    if (
+      !isLoading &&
+      !isFetched &&
+      enableAutoFetch &&
```
boliveira, line 58: If this effect will only create a exchange filter, then this should be called `enableAutoCreate`. But then the hook itself should be called `useCreateExchangeFilter` and the `isFetched` value should be changed to something like `isCreated`. I think this needs to be thought out a little bit more. Let's talk tomorrow about this.

packages/react/src/exchanges/hooks/types/useExchange.ts
andreecosta: Fixed!

packages/react/src/exchanges/hooks/types/useExchange.ts
```diff
@@ -0,0 +1,14 @@
...
+  exchangeBookRequestId?: ExchangeBookRequest['id'];
```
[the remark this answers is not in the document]
andreecosta: Fixed.

packages/react/src/exchanges/hooks/types/useExchangeBookRequest.ts
andreecosta: Done.

packages/react/src/exchanges/hooks/useExchange.ts
andreecosta: Fixed.

packages/react/src/exchanges/hooks/useExchangeFilter.ts
andreecosta: And done. I think I've changed everything we talked yesterday, but it you notice something missing, please let me know.

boliveira (approved):

jpklzm (approved):

pedro-gbf (approved):

PR #804, opened 2023-06-27
feat: setup size preferences redux structure

talbertosilva:
## Description

- Setup size preferences redux structure which fetches /listingFacets endpoint.
- Added the respective tests

### Dependencies

None

## Checklist

## Disclaimer


packages/redux/src/sizePreferences/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+      expect(state).toEqual(initialState.error);
+      expect(state).toBeNull();
+    });
+
+    it('should handle FETCH_LISTING_FACETS_REQUEST action type', () => {
+      const expectedResult = initialState.error;
+
+      expect(
+        reducer(undefined, {
```
boliveira, line 40: Please, do not use `undefined` in these tests as this is already covered by the `should return the initial state` test. Use a state that is representative for when this action happens (like isLoading: false and error set to a value).

packages/redux/src/sizePreferences/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+describe('sizePreferences reducer', () => {
+  beforeEach(() => {
+    initialState = fromReducer.INITIAL_STATE;
+  });
+
+  describe('reset handling', () => {
+    it('should return the initial state', () => {
+      expect(
+        reducer(undefined, {
```
boliveira, line 21: Use a different value instead of undefined for the state so that the action is properly tested.

packages/redux/src/sizePreferences/selectors.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { StoreState } from '../types/index.js';
+
+/**
+ * Returns the sizePreferences error.
+ *
+ * @param state - Application state.
+ *
+ * @returns Error.
+ */
+export const getSizePreferencesError = (state: StoreState) =>
```
boliveira, line 12: The action is called `fetchListingFacets` but the selector does not match that name. Either all clients, actions and selectors follow the naming listingFacets or sizePreferences. In my opinion, sizePreferences is a much better name but check if that is possible. If you do choose to go with the sizePreferences, then please change every name that still refers to listingFacets to use that name (like fetchSizePreferences, getSizePreferences, etc...)

talbertosilva: I can see that using those two names might be confusing and I will change to one of those two. Choosing listingFacets might be the most accurate since the endpoint is retrieving listing facets, what do you think?

boliveira: I am not aware of what is the intended usage of this endpoint but if the idea is to give users a way to find the size preferences, then I would not call it listingfacets even though the endpoint is called like that. The name of our clients can be different than the name of the resources depending if we want to add an abstraction over them which will facilitate the consumers understanding. Talk to your team members about this as the name will depend on what feature are we intended to give to users with this change.

packages/client/src/products/types/productListingFacets.types.ts
```diff
@@ -0,0 +1,26 @@
...
+  slug: string;
+  url: string;
+  parentId: number;
+  _isDisabled: boolean;
+  _isActive: boolean;
+  groupsOn: number;
+  count: number;
+};
+
+export type ProductListingFacet = {
```
boliveira, line 14: What is the difference between this type and `FacetGroup` type that represents the facets that comes on the get product listing response? I think they are the same thing. Can you check, please?

packages/redux/src/products/selectors/listings.ts
```diff
@@ -760,3 +761,49 @@ export const getProductListingFacetGroups = createSelector(
...
+  getListingFacetsState((state.products as ProductsState).lists).result;
+
+/**
+ * Returns the loading status for the listing facets.
+ *
+ * @param state - Application state.
+ *
+ * @returns Listing facets loading status.
+ */
+export const isListingFacetsLoading = (state: StoreState) =>
```
boliveira, line 808:
```suggestion
export const areListingFacetsLoading = (state: StoreState) =>
```

packages/redux/src/products/selectors/listings.ts
```diff
@@ -760,3 +761,49 @@ export const getProductListingFacetGroups = createSelector(
...
+ *
+ * @param state - Application state.
+ *
+ * @returns Listing Facets error.
+ */
+export const getListingFacetsError = (state: StoreState) =>
+  getListingFacetsState((state.products as ProductsState).lists).error;
+
+/**
+ * Returns the result of the listing facets.
```
boliveira, line 792:
```suggestion
 * Returns the result of the fetch listing facets request.
```

The same for other selectors

packages/client/src/products/types/facetValue.types.ts
```diff
@@ -6,6 +6,7 @@ export type FacetValue = {
   url: string;
   parentId: number;
   groupsOn: number;
+  groupType?: number;
```
boliveira, line 9: Is this correct? I think this is only being added to the FacetEntity by the facets schema by the fetchProductListingFacetsFactory module.

packages/redux/src/products/selectors/listings.ts
```diff
@@ -760,3 +761,49 @@ export const getProductListingFacetGroups = createSelector(
...
+
+/**
+ * Returns all facets by parent type.
+ *
+ * @param state - Application state.
+ * @param facetGroupType - Facet group type to find.
+ *
+ * @returns Array with all facets content requested by group type.
+ */
+export const getFacetsByGroupType = (
```
boliveira, line 773: I think it is better to use a `createSelector` from `reselect` here to avoid returning new instances even when the values haven't changed, which will avoid consumers to find out if there was a change or not between two consecutive calls to this selector. Check other selectors that use `createSelector` to know how to do this.

talbertosilva: Done!

packages/client/src/products/types/facetValue.types.ts
talbertosilva: Indeed, this parameter addition wasn't needed, thanks, fixed!

packages/redux/src/products/selectors/listings.ts
talbertosilva: Fixed!

packages/redux/src/products/selectors/listings.ts
talbertosilva: Fixed!

boliveira (approved):

ruifcnunes (approved):

PR #833, opened 2023-08-01
feat(react) useUserContact and useUserContacts hooks

pedro-gbf:
## Description

This adds both new hooks for user contacts management, one for using with a single address and the other with the user contacts list.

### Dependencies

None

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/users/hooks/types/useUserContacts.types.ts
+++ b/packages/react/src/users/hooks/types/useUserContacts.types.ts
@@ -1,0 +1,6 @@
+import type { Config } from '@farfetch/blackout-client';
+
+export type UseUserContactsOptions = {
+  enableAutoFetch?: boolean;
+  fetchConfig?: Config;
+};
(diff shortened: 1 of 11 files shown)


packages/react/src/users/hooks/types/useUserContacts.types.ts
```diff
@@ -0,0 +1,6 @@
+import type { Config } from '@farfetch/blackout-client';
+
+export type UseUserContactsOptions = {
```
boliveira, line 3: I think it is better to have one options type for each hook as they might diverge in the future.

boliveira (approved):

jpklzm (approved):

packages/react/src/users/hooks/types/useUserContacts.types.ts
pedro-gbf: 👍  Changed it

boliveira (approved):

ruifcnunes (approved):

PR #836, opened 2023-08-03
feat(redux): add shared wishlist middleware for updating snapshop

danpadrinom:
## Description

- This adds the shared wishlist middleware that handles the update of the snapshot when a wishlist item is updated, added or deleted.

### Dependencies

None

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/sharedWishlists/middlewares/__tests__/updateSharedWishlistUponItemsChanges.test.ts
+++ b/packages/redux/src/sharedWishlists/middlewares/__tests__/updateSharedWishlistUponItemsChanges.test.ts
@@ -1,0 +1,70 @@
+import * as actionTypes from '../../../wishlists/actionTypes.js';
+import { INITIAL_STATE } from '../../reducer.js';
+import {
+  mockSharedWishlistId,
+  mockSharedWishlistState,
+} from 'tests/__fixtures__/sharedWishlists/sharedWishlists.fixtures.mjs';
+import { mockStore } from '../../../../tests/index.js';
+import { updateSharedWishlist } from '../../actions/index.js';
+import thunk from 'redux-thunk';
+import updateSharedWishlistUponItemsChanges from '../updateSharedWishlistUponItemsChanges.js';
+import type { UserEntity } from '../../../index.js';
+
+jest.mock('../../actions', () => ({
+  ...jest.requireActual('../../actions'),
+  updateSharedWishlist: jest.fn(() => () => ({ type: 'foo' })),
+}));
+
+describe('updateSharedWishlistUponItemsChanges', () => {
+  it('should do nothing if the action is not adding, deleting or updating a wishlist item', () => {
+    const store = mockStore({ sharedWishlist: INITIAL_STATE }, {}, [
+      updateSharedWishlistUponItemsChanges,
+    ]);
+
+    store.dispatch({ type: 'foo' });
+
+    expect(updateSharedWishlist).not.toHaveBeenCalled();
+  });
+
+  it.each([
+    actionTypes.ADD_WISHLIST_ITEM_SUCCESS,
+    actionTypes.REMOVE_WISHLIST_ITEM_SUCCESS,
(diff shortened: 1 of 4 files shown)


packages/redux/src/sharedWishlists/middlewares/updateSharedWishlistUponItemsChanges.ts
```diff
@@ -0,0 +1,56 @@
...
+  >;
+  // Returns the current redux state.
+  getState: () => StoreState;
+};
+
+const updateSharedWishlistUponItemsChanges: Middleware =
+  ({ dispatch, getState }: UpdateSharedWishlistUponItemsChangesParams) =>
+  next =>
+  (
+    action:
```
boliveira, line 33: `action` here can be any action, not only those related to wishlists. The correct type should be `AnyAction` from redux.

packages/redux/src/sharedWishlists/middlewares/__tests__/updateSharedWishlistUponItemsChanges.test.ts
```diff
@@ -0,0 +1,67 @@
...
+          },
+        },
+        [thunk, updateSharedWishlistUponItemsChanges],
+      );
+
+      store.dispatch({
+        type: actionType,
+      });
+
+      expect(updateSharedWishlist).toHaveBeenCalled();
```
boliveira, line 64: Check if it is calling with the correct shared wishlist id.

packages/redux/src/sharedWishlists/middlewares/index.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Shared wishlists middlewares.
+ */
+export { default as updateSharedWishlistUponItemsChanges } from './updateSharedWishlistUponItemsChanges.js';
```
boliveira, line 4: You will need to re-export this in the `packages/redux/src/sharedWishlists/index.ts` file so that it can be available for the consumers to import from the root of the package.

packages/redux/src/sharedWishlists/middlewares/updateSharedWishlistUponItemsChanges.ts
danpadrinom: Updated ✅

packages/redux/src/sharedWishlists/middlewares/__tests__/updateSharedWishlistUponItemsChanges.test.ts
danpadrinom: Great idea, done!

packages/redux/src/sharedWishlists/middlewares/index.ts
danpadrinom: Right! Done.

packages/redux/src/sharedWishlists/middlewares/updateSharedWishlistUponItemsChanges.ts
```diff
@@ -0,0 +1,45 @@
...
+const updateSharedWishlistUponItemsChanges: Middleware =
+  ({ dispatch, getState }: UpdateSharedWishlistUponItemsChangesParams) =>
+  next =>
+  (action: AnyAction) => {
+    if (
+      action.type === actionTypes.UPDATE_WISHLIST_ITEM_SUCCESS ||
+      action.type === actionTypes.ADD_WISHLIST_ITEM_SUCCESS ||
+      action.type === actionTypes.REMOVE_WISHLIST_ITEM_SUCCESS
+    ) {
+      const user = getUser(getState());
```
ruifcnunes, line 32: What do you think of storing the `state` inside a const here? that way we would not have to call `getState()` function twice (here and on line 36).

danpadrinom: Agree! Updated

ruifcnunes (approved):

PR #845, opened 2023-08-18
fix: component thumbnail index export

vinicaval:
## Description

Fix the thumbnail index export, the extension of file was wrong.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/contents/components/base/media/components/thumbnail/index.ts
+++ b/packages/react/src/contents/components/base/media/components/thumbnail/index.ts
@@ -1,1 +1,1 @@
-export { default } from './Thumbnail.js';
+export { default } from './Thumbnail.jsx';


packages/react/src/contents/components/base/media/components/thumbnail/index.ts
```diff
@@ -1 +1 @@
-export { default } from './Thumbnail.js';
+export { default } from './Thumbnail.jsx';
```
richard190m, line 1: The correct file extension is .tsx

richard190m (changes requested):

danielbento92 (approved):

packages/react/src/contents/components/base/media/components/thumbnail/index.ts
danielbento92: I think it's jsx.
It's like other index folders. If you check like `packages/react/src/bags/hooks/index.ts`, was exporting from useBag.js, but our file is useBag.ts.

I think it's something related with some configuration.

When it builds, it generate some files. In these files, the TSX file, results in JSX file. The index, should point to this result file, which should be "Thumbnail.jsx"

richard190m: Yes, you're right. I will approve the PR

richard190m (approved):

PR #847, opened 2023-08-22
feat(core): add middleware for getting the wishlist sets upon shared wishlist id changes

danpadrinom:
## Description

This adds a middleware for getting the wishlist sets when the `sharedWishlistId` has changed. It ensures we are always accessing the updated data from the wishlist sets after creating or deleting a shared wishlist.

Also, the action `UPDATE_WISHLIST_SET_SUCCESS` was added to the `updateSharedWishlistUponItemsChanges` middleware and the shared wishlist reducer (result) was adjusted.

### Dependencies

None

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/sharedWishlists/redux/__tests__/reducer.test.js
+++ b/packages/core/src/sharedWishlists/redux/__tests__/reducer.test.js
@@ -270,6 +270,14 @@ describe('shared wishlists reducer', () => {
       ).toBe(expectedResult);
     });
 
+    it('should handle REMOVE_SHARED_WISHLIST_SUCCESS action type', () => {
+      expect(
+        reducer(currentState, {
+          type: actionTypes.REMOVE_SHARED_WISHLIST_SUCCESS,
+        }).result,
+      ).toBe(initialState.result);
+    });
+
     it('should handle other actions by returning the previous state', () => {
       const state = { result: 'foo' };
 
--- a/packages/core/src/sharedWishlists/redux/middlewares/getWishlistSetsUponSharedWishlistIdChanges.js
+++ b/packages/core/src/sharedWishlists/redux/middlewares/getWishlistSetsUponSharedWishlistIdChanges.js
@@ -1,0 +1,35 @@
+import {
+  CREATE_SHARED_WISHLIST_SUCCESS,
+  REMOVE_SHARED_WISHLIST_SUCCESS,
+} from '../actionTypes';
+import { doGetWishlistSets } from '../../../wishlists/redux/actions';
+import { getUser, getUserIsGuest } from '../../../entities/redux/selectors';
+import { getWishlistsSets as getWishlistsSetsClient } from '../../../wishlists/client';
+
+const getWishlistSets = doGetWishlistSets(getWishlistsSetsClient);
+
+/**
+ * Middleware to get the updated wishlist sets if a shared wishlist has been created or deleted.
+ *
+ * @function getWishlistSetsUponSharedWishlistIdChanges
+ * @memberof module:sharedWishlists/middlewares
+ *
+ * @param {object} store - Redux store at the moment.
+ *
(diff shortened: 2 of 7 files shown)


packages/core/src/sharedWishlists/redux/middlewares/getWishlistSetsUponSharedWishlistIdChanges.js
```diff
@@ -0,0 +1,35 @@
...
+ * @function getWishlistSetsUponSharedWishlistIdChanges
+ * @memberof module:sharedWishlists/middlewares
+ *
+ * @param {object} store - Redux store at the moment.
+ *
+ * @returns {Function} Redux middleware.
+ */
+export default store => next => action => {
+  if (
+    action.type === REMOVE_SHARED_WISHLIST_SUCCESS ||
```
boliveira, line 23: Just to understand: If a shared wishlist is removed, why do we need to refetch the wishlist sets?

boliveira:
Also there is a problem I have just noticed in the implementation, at least in `main` and I believe `v1` also has this problem. In `packages/redux/src/sharedWishlists/reducer.ts` when the `REMOVE_SHARED_WISHLIST_SUCCESS` is dispatched, the `result` slice is not updated which means the `getSharedWishlistId` selector will keep returning the previous shared wishlist id. Can you take the opportunity to take a look at that case as well, please? If you prefer, you can have it in another PR but that should be tackled. Thank you!

packages/core/src/sharedWishlists/redux/middlewares/getWishlistSetsUponSharedWishlistIdChanges.js
danpadrinom: We need to refetch when a shared wishlist is removed in case we want to immediately share it again. This way we warrantee we have the latest value for the `sharedWishlistId` in the wishlist set.

danpadrinom:
The scope of this PR has been changed as it now adds an action to the `updateSharedWishlistUponItemsChanges` middleware.

packages/core/src/sharedWishlists/redux/__tests__/reducer.test.js
```diff
@@ -270,6 +270,15 @@ describe('shared wishlists reducer', () => {
       ).toBe(expectedResult);
     });
 
+    it.only('should handle REMOVE_SHARED_WISHLIST_SUCCESS action type', () => {
```
boliveira, line 273:
```suggestion
    it('should handle REMOVE_SHARED_WISHLIST_SUCCESS action type', () => {
```

packages/core/src/sharedWishlists/redux/__tests__/reducer.test.js
```diff
@@ -270,6 +270,15 @@ describe('shared wishlists reducer', () => {
...
+      expect(
+        reducer(currentState, {
+          payload: { result: 'foo' },
```
boliveira, line 276: Also, the action will not contain a `payload` property when dispatched by the `doDeleteSharedWishlist.js` action.

danpadrinom: Right! Updated

packages/core/src/sharedWishlists/redux/__tests__/reducer.test.js
danpadrinom: Updated 👍

packages/core/src/sharedWishlists/redux/middlewares/getWishlistSetsUponSharedWishlistIdChanges.js
boliveira: Ok, thanks!

boliveira (approved):

PR #874, opened 2023-10-12
feat(analytics|react): update recommendations tracking parameters

talbertosilva:
## Description

- Updated recommendations tracking parameters (additions and naming changes)
- Updated unit tests related to those same changes

### Dependencies

None.

## Checklist

## Disclaimer


packages/analytics/src/integrations/Omnitracking/definitions.ts
```diff
@@ -600,20 +603,27 @@ export const trackEventsMapper: Readonly<OmnitrackingTrackEventsMapper> = {
     actionArea: data.properties?.from,
     productId: getProductId(data.properties),
     lineItems: getProductLineItems(data),
+    moduleTitle: JSON.stringify([data.properties?.list]),
```
danielbento92, line 606: To avoid these duplications of moduleTitle and Id, would be more secure create an helper to deal with this module, then call the helper function in each event.

In this method you can also check if this property are undefined after JSON.stringify to avoid "[undefined]" values.

Also, check what would be the correct behaviour in undefined properties situation (if fill with empty array, or if assign the property with undefined instead of empty array)

danielbento92 (changes requested):

packages/analytics/src/integrations/Omnitracking/definitions.ts
talbertosilva: Fixed!
Created a helper function to avoid duplications, but also check if the parameters are undefined and check if it's a non recommendations product details event (in this scenario it's not necessary to send these parameters)!
Data team agreed that in case _moduleId_ or _moduleTitle_ are undefined, it should be sent an empty array

packages/analytics/src/integrations/Omnitracking/__tests__/omnitracking-helper.test.ts
```diff
@@ -237,3 +243,63 @@ describe('getGenderValueFromProperties', () => {
...
+  it('should return the json stringified values correctly', () => {
+    expect(
+      getRecommendationsTrackingData({
+        event: 'abc',
+        properties: {
+          from: FromParameterType.Recommendations,
+          ...mockedRecommendationsValues,
+        } as Record<string, unknown>,
+      } as EventData<TrackTypesValues>),
+    ).toEqual({ ...mockedReturnedRecommendationsValues });
```
danielbento92, line 289: Spread not needed.

`expect(
       getRecommendationsTrackingData({
         event: 'abc',
         properties: {
           from: FromParameterType.Recommendations,
           ...mockedRecommendationsValues,
         } as Record<string, unknown>,
       } as EventData<TrackTypesValues>),
     ).toEqual(mockedReturnedRecommendationsValues);`

talbertosilva: Fixed 👍🏼

boliveira (approved):

danielbento92 (approved):

PR #877, opened 2023-10-16
fix: fix hooks multi requests

jpklzm:
## Description
This adds a new validation layer on hooks to prevent multiple requests.

## Checklist

## Disclaimer


packages/react/src/bags/hooks/useBag.ts
```diff
@@ -159,7 +160,16 @@ const useBag = (options: UseBagOptions = {}) => {
   );
 
   useEffect(() => {
-    if (!isLoading && !isFetched && enableAutoFetch && userBagId) {
+    const updatedState = store.getState();
+    const updatedIsLoading = isBagLoading(updatedState as StoreState);
```
boliveira, line 164: Instead of casting here, cast the `updatedState` variable to `StoreState` type like this:

```suggestion
    const updatedState = store.getState() as StoreState;
    const updatedIsLoading = isBagLoading(updatedState);
    const updatedIsFetched = isBagFetched(updatedState);
```

boliveira: The same applies to the other hooks

boliveira (approved):

packages/react/src/bags/hooks/useBag.ts
```diff
@@ -159,7 +160,16 @@ const useBag = (options: UseBagOptions = {}) => {
   );
 
   useEffect(() => {
-    if (!isLoading && !isFetched && enableAutoFetch && userBagId) {
+    const updatedState = store.getState() as StoreState;
```
diegogriep, line 163: Is this type not inferred by [getEntities](https://github.com/Farfetch/blackout/blob/main/packages/redux/src/entities/selectors/entity.ts#L43C5-L43C5)?
My concern is to avoid repeating `as StoreState` on every store.getState() call.

boliveira: `store.getState()` will return a value of the type that was passed to the `useStore()` call. As we are not passing any there it will assume as unknown, so we have to add a cast here. Both approaches are ok, but specifying the type on the `useStore()` might be cleaner but for this specific case we can leave as it is. What do you think?

dnamorim (approved):

packages/react/src/bags/hooks/useBag.ts
diegogriep: For sure. We can improve the generics in the future as well.

diegogriep (approved):

ruifcnunes (approved):

PR #881, opened 2023-10-18
fix(redux): avoid facets replacement when multiple listing requests

danpadrinom:
## Description

This adjusts the `facet` and `productListing` schemas for avoiding the replacement of facets when multiple product listing requests are made.

**IMPORTANT:** BREAKING CHANGE TBD.

### Dependencies

None.

## Checklist

## Disclaimer


packages/redux/src/entities/schemas/facet.ts
```diff
@@ -7,9 +7,9 @@ import { schema } from 'normalizr';
 
 export const getId = (
   { value, valueUpperBound, groupsOn }: FacetValue,
-  { key, type: parentType }: FacetGroup,
+  { key, type: parentType, hash }: FacetGroup,
 ) =>
-  `${key}_${value}${
+  `${key}_${value}_${hash}${
```
boliveira, line 12: The `hash` portion of the id must go at the end. If the facet is for a size or sizes by category for example, it will be inserted before the groupsOn portion.

packages/client/src/products/types/facetGroup.types.ts
```diff
@@ -57,4 +57,5 @@ export type FacetGroup = {
   _isClearHidden: boolean;
   _isClosed: boolean;
   dynamic: number;
+  hash: string;
```
boliveira, line 60: The hash will only be added to the FacetGroup entity in redux. This type is defined in the client package which can be used without redux. You would need to change the FacetGroup type inside the listing to account for that but I think we should not expose the hash to the consumers as it is an implementation detail.

packages/redux/src/entities/schemas/productsList.ts
```diff
@@ -97,8 +105,9 @@ export default new schema.Entity(
         },
       );
       const productsList = {
-        facetGroups,
+        facetGroups: newFacetGroups,
         filterSegments: newFilterSegments,
+        hash,
```
boliveira, line 110: This can be removed as it is not necessary for normalisation or consumption from the clients.

packages/redux/src/entities/schemas/facet.ts
danpadrinom: Right! Updated

packages/redux/src/entities/schemas/productsList.ts
danpadrinom: If we don't return the hash here, then we get an issue in the boilerplate at the `getListingActiveFiltersByFacetGroup ` 🤔

packages/client/src/products/types/facetGroup.types.ts
danpadrinom: Agree 👍 updated!

packages/redux/src/entities/schemas/productsList.ts
boliveira: You are right, I thought this was not necessary for clients to access it.

packages/redux/src/entities/types/productListing.types.ts
```diff
@@ -7,11 +7,12 @@ import type {
...
 import type { ProductEntity } from './product.types.js';
 
-export type FacetGroupsNormalized = Array<
-  Omit<FacetGroup, 'values'> & {
-    values: FacetEntity['id'][][];
-  }
->;
+export type FacetGroupNormalized = Omit<FacetGroup, 'values'> & {
+  values: FacetEntity['id'][][];
+  hash?: string;
```
boliveira, line 12: I think this should be removed since it should not be consumed by clients, even though that property exists. Did you need to use it in boilerplate?

packages/redux/src/entities/schemas/facet.ts
```diff
@@ -4,17 +4,25 @@ import {
...
 import { schema } from 'normalizr';
+import type { FacetGroupNormalized } from '../index.js';
 
 export const getId = (
-  { value, valueUpperBound, groupsOn }: FacetValue,
-  { key, type: parentType }: FacetGroup,
-) =>
-  `${key}_${value}${
+  entity: FacetValue,
+  parent: FacetGroup | FacetGroupNormalized,
```
boliveira, line 11: `parent` will always be a `FacetGroup` with a `hash` property (if you do some changes that I will describe shortly) as `getId` is only called when applying normalisation to a facet so it should be this:

```suggestion
  parent: FacetGroupWithListingHash,
```

where `FacetGroupWithListingHash` is this:

```suggestion
export type FacetGroupWithListingHash = FacetGroup & { hash?: string };
```

Define that type in this file as the suggestion as it will be useful in `productsList` schema.

packages/redux/src/entities/schemas/productsList.ts
```diff
@@ -28,6 +28,7 @@ export default new schema.Entity(
       const {
```
boliveira, line 28: Here you will need to do the following:

In line 53, replace `facetGroups` with `newFacetGroups` so the `facetId` property of the filterSegments are calculated correctly with the hash. This will help finding the corresponding facet. 

In lines 89 and 102, replace the cast to `FacetGroup` to `FacetGroupWithListingHash` (that is the type that I suggested on the comment for the facet entity schema which you will need to import)

packages/redux/src/entities/types/productListing.types.ts
```diff
@@ -7,8 +7,10 @@ import type {
 } from '@farfetch/blackout-client';
 import type { ProductEntity } from './product.types.js';
 
+export type FacetGroupWithListingHash = FacetGroup & { hash?: string };
+
 export type FacetGroupsNormalized = Array<
-  Omit<FacetGroup, 'values'> & {
+  Omit<FacetGroupWithListingHash, 'values'> & {
```
boliveira, line 13: Keep FacetGroup here as I do not think we should expose the hash property to consumers yet.

packages/redux/src/entities/schemas/productsList.ts
danpadrinom: Updated, thanks!

packages/redux/src/entities/schemas/facet.ts
danpadrinom: Great, updated!

packages/redux/src/entities/types/productListing.types.ts
danpadrinom: Right, I exposed it here and also in the `FacetEntity`, but it's not needed. Fixed!

packages/redux/src/entities/types/productListing.types.ts
danpadrinom: I thought it was needed, but a better approach was taken from the boilerplate side! Removed.

giuliamml (approved):

boliveira (approved):

ruifcnunes (approved):

PR #890, opened 2023-10-25
feat: add isMainWishlist parameter to `main`

talbertosilva:
## Description

- Add isMainWishlist parameter to wishlist related events
- Updated unit tests as well

### Dependencies

None.

## Checklist

## Disclaimer


danielbento92 (approved):

ruifcnunes:
I believe we should also update the GA4 schemas for wishlist related events with this new field, even if it is optional (we can at least define its type and state that it is not required). 

What do you think?

ruifcnunes (approved):

PR #903, opened 2023-11-06
feat: add unitSalePrice to analytics events `v1`

talbertosilva:
## Description

- Add unitSalePrice to lineItems and product_details event.
- Add a helper to calculate unitSalePrice (price - discount).
- Updated snapshots

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/analytics/integrations/Omnitracking/omnitracking-helper.js
+++ b/packages/core/src/analytics/integrations/Omnitracking/omnitracking-helper.js
@@ -322,6 +322,28 @@ export const getValParameterForEvent = (valParameters = {}) => {
 };
 
 /**
+ * Get the product unit sale price from a specific product.
+ *
+ * @param {object} productData - Product data.
+ *
+ * @returns {number|undefined} The unit sale price value, or undefined if no pricing data was found in productData argument.
+ */
+export const getProductUnitSalePrice = productData => {
+  if (typeof productData.unitSalePrice === 'number') {
+    return productData.unitSalePrice;
+  }
+
+  if (
+    typeof productData.priceWithoutDiscount === 'number' &&
+    typeof productData.discountValue === 'number'
+  ) {
+    return productData.priceWithoutDiscount - productData.discountValue;
+  }
+
+  return undefined;
+};
+
+/**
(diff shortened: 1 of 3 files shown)


packages/core/src/analytics/integrations/Omnitracking/omnitracking-helper.js
```diff
@@ -321,6 +321,25 @@ export const getValParameterForEvent = (valParameters = {}) => {
   return JSON.stringify(valParameters);
 };
 
+/**
+ * Get the product unit sale price from a specific product.
+ *
+ * @param {object} productData - Product data.
+ *
+ * @returns {*} The unit sale price value.
```
boliveira, line 329: This should be something like this since this function returns a number or undefined:
```suggestion
 * @returns {number|undefined} The unit sale price value, or undefined if no pricing data was found in productData argument.
```

packages/core/src/analytics/integrations/Omnitracking/omnitracking-helper.js
```diff
@@ -321,6 +321,25 @@ export const getValParameterForEvent = (valParameters = {}) => {
...
+ */
+export const getProductUnitSalePrice = productData => {
+  if (typeof productData.unitSalePrice === 'number') {
+    return productData.unitSalePrice;
+  }
+
+  if (productData.priceWithoutDiscount && productData.discountValue) {
```
boliveira, line 336: I know this is not possible to happen but if either priceWithoutDiscount or discountValue are 0, the function will return undefined. I think that to be on the safe side we should just check if both are number.

packages/core/src/analytics/integrations/Omnitracking/omnitracking-helper.js
talbertosilva: Fixed 👍🏼

packages/core/src/analytics/integrations/Omnitracking/omnitracking-helper.js
talbertosilva: Fixed 👍🏼

danielbento92 (approved):

danielbento92:
Rebase the branch before merge.

boliveira (approved):

PR #905, opened 2023-11-07
fix: commerce pages merge strategy not merging correctly

richard190m:
## Description

* Fix commerce pages merge strategy to be able to merge custom and base component, with the new structure with sections and subsections

## Checklist

## Disclaimer


joaoprcosta (approved):

packages/redux/src/contents/utils.ts
```diff
@@ -25,6 +25,40 @@ export enum ContentEnvironmentCode {
...
+ */
+const defaultContentReturn = {
+  number: 0,
+  totalItems: 0,
+  totalPages: 0,
+  entries: [],
+};
+
+/**
+ * Constant that represent the default section object returned by content api
```
boliveira, line 39: This is a function so I would rephrase this to something like: "Function that returns the default section which would be returned by content api containing content generated by Content Tool"

packages/redux/src/contents/utils.ts
```diff
@@ -162,7 +196,7 @@ export const getBestRankedCommercePageUsingDefaultStrategy = (
 
 /**
  * Method to return the best ranked commerce page with all components from other
- * pages without duplicates.
+ * pages without duplicates. This implementation only works with content from new Content Tool
```
boliveira, line 199: Question: Will this be a breaking change for current consumers of this function?

richard190m: This implementation is not used by any tenant. Currently, the only tenant that uses the merge strategy is Harrods, and they have their own implementation in their bff. This is a fix to BG implement merge strategy using blackout now

packages/redux/src/contents/utils.ts
richard190m: Fixed. Thanks

boliveira (approved):

PR #908, opened 2023-11-07
fix: save exchange filters by id

andreecosta:
## Description

BREAKING CHANGE:
- Exchange filters are now stored under entities by orderItemUuid (shippingOrderLineId).
Previously the user wasn't able to fetch multiple filters because each request cleared
the previous data, which this PR aims to fix.
- useCreateExchangeFilter hook renamed to useExchangeFilters.
- isCreating selector from the hook, renamed to isLoading.

## Checklist

## Disclaimer


packages/react/src/exchanges/hooks/useCreateExchangeFilter.ts
```diff
@@ -36,27 +51,30 @@ function useCreateExchangeFilter() {
...
 
   const data = useMemo(() => {
-    if (!exchangeFilterResult) {
+    if (!exchangeFilters && !exchangeFilterById) {
       return undefined;
     }
 
-    return exchangeFilterResult;
-  }, [exchangeFilterResult]);
+    return { exchangeFilters, exchangeFilterById };
```
boliveira, line 68: Why the `useCreateExchangeFilter` needs to return all exchange filters? This should return only the new exchange filter. If the user needs to have all exchange filters, then we should have a hook called `useExchangeFilters` and one of its actions should be `create`.

packages/react/src/exchanges/hooks/useExchange.ts
```diff
@@ -93,7 +93,7 @@ function useExchange(
 
   const {
     actions: { create: createExchangeFilter },
-  } = useExchangeFilter();
+  } = useExchangeFilter(orderItemUuid);
```
boliveira, line 96: Please, change the name of this import. It is being imported as `useExchangeFilter` but it is being imported from `useCreateExchangeFilter`. Also, if orderItemUuid is optional in `useExchange`, it should be optional in `useCreateExchangeFilter` as well. You can add an optional `orderItemUuid` parameter to the create action from `useCreateExchangeFilter` so that the user can set/override the orderItemUuid parameter from the hook.

packages/redux/src/exchanges/actions/resetExchangeFiltersState.ts
```diff
@@ -0,0 +1,53 @@
+import * as actionTypes from '../actionTypes.js';
```
boliveira, line 1: Since we are adding a breaking change, rename this file to `resetExchangeFilters` and change the export in `packages/redux/src/exchanges/actions/index.ts` to `resetExchangeFilters` as well.

packages/react/src/exchanges/hooks/useCreateExchangeFilter.ts
```diff
@@ -36,27 +51,30 @@ function useCreateExchangeFilter() {
...
+  }, [exchangeFilters, exchangeFilterById]);
 
   return {
     isCreating,
     error,
     data,
     actions: {
       create,
+      reset,
```
boliveira, line 77: The `reset` action also does not make sense to be exported here since it will reset all state and not the one that was resulted from the user invocation of the hook.

packages/react/src/exchanges/hooks/useCreateExchangeFilter.ts
andreecosta: We talked in private and decided to rename this hook and change its behaviour. It returns all exchange filters only.

packages/react/src/exchanges/hooks/useExchange.ts
andreecosta: Fixed

packages/redux/src/exchanges/actions/resetExchangeFiltersState.ts
andreecosta: Done 👍

packages/react/src/exchanges/hooks/useExchange.ts
```diff
@@ -27,7 +27,7 @@ function useExchange(
   options: UseExchangeOptions = {},
 ) {
   const exchangeIdHookParameter = exchangeId;
-  const { enableAutoFetch = true, fetchConfig } = options;
+  const { enableAutoFetch = true, fetchConfig, orderItemUuid = '' } = options;
```
boliveira, line 30: Since `useExchangeFilters` accepts a undefined value for orderItemUuid, you do not need to provide a default value for it here.

```suggestion
  const { enableAutoFetch = true, fetchConfig, orderItemUuid } = options;
```

packages/react/src/exchanges/hooks/useExchangeFilters.ts
```diff
@@ -0,0 +1,111 @@
...
+import { useSelector } from 'react-redux';
+import useAction from '../../helpers/useAction.js';
+import type {
+  Config,
+  ExchangeFilterItem,
+  OrderItem,
+  PostExchangeFilterData,
+} from '@farfetch/blackout-client';
+
+type PostExchangeFilterDataHook = PostExchangeFilterData & {
```
boliveira, line 19: This type must be exported since they are part of a public method (This applies to `ExchangeFilterItemWithOptionalUuid` as well)

packages/react/src/exchanges/hooks/useExchangeFilters.ts
```diff
@@ -0,0 +1,111 @@
...
+
+type ExchangeFilterItemWithOptionalUuid = Omit<
+  ExchangeFilterItem,
+  'orderItemUuid'
+> & {
+  orderItemUuid?: OrderItem['shippingOrderLineId'];
+};
+
+/**
+ * Provides facilities to create an exchange filter.
```
boliveira, line 31:
```suggestion
 * Provides facilities to manage exchange filters.
```

packages/react/src/exchanges/hooks/useExchangeFilters.ts
```diff
@@ -0,0 +1,111 @@
...
+      ? getExchangeFilterError(state, orderItemUuidHookParameter)
+      : null,
+  );
+  const exchangeFilters = useSelector(getExchangeFilters);
+  const createExchangeFilter = useAction(createExchangeFilterAction);
+  const reset = useAction(resetExchangeFilters);
+
+  /**
+   * Creates an exchange filter. This method will save the exchange filter by orderItemUuid without
+   * replacing existing filters. Will throw if no exchangeFilterData is passed
```
boliveira, line 51:
```suggestion
   * replacing existing filters. Will throw in the following conditions:
     - exchangeFilterData parameter is not passed;
     - or no usable `orderItemUuid` is found. If it is not passed in `exchangeFilterData`, the one passed on the hook call will be used.
```

packages/react/src/exchanges/hooks/useExchangeFilters.ts
```diff
@@ -0,0 +1,111 @@
...
+
+      if (!orderItemUuid && !orderItemUuidHookParameter) {
+        return Promise.reject(
+          new Error(
+            'No orderItemUuid provided to either the hook or the create function',
+          ),
+        );
+      }
+
+      exchangeFilterData.exchangeFilterItems.forEach(item => {
```
boliveira, line 78: This can be simplified by checking beforehand if the `orderItemUuid` is undefined:

```suggestion
      if (!orderItemUuid) {
          exchangeFilterData.exchangeFilterItems.forEach(item => {
            item.orderItemUuid = orderItemUuidHookParameter;
      ...
```

packages/react/src/exchanges/hooks/useExchangeFilters.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+      });
+
+      return createExchangeFilter(exchangeFilterData, config);
+    },
+    [createExchangeFilter, orderItemUuidHookParameter],
+  );
+
+  const data = useMemo(() => {
+    if (!exchangeFilters) {
```
boliveira, line 93: This is unnecessary, you can just return the exchange filters like this as it will have the same meaning:

```suggestion
   return exchangeFilters;
```

packages/redux/src/exchanges/__tests__/reducer.test.ts
```diff
@@ -399,8 +401,8 @@ describe('exchanges reducer', () => {
 
   describe('getExchangeFilter() selector', () => {
```
boliveira, line 402:
```suggestion
  describe('getExchangeFilters() selector', () => {
```

packages/react/src/exchanges/hooks/useExchangeFilters.ts
```diff
@@ -0,0 +1,111 @@
...
+> & {
+  orderItemUuid?: OrderItem['shippingOrderLineId'];
+};
+
+/**
+ * Provides facilities to create an exchange filter.
+ */
+function useExchangeFilters(orderItemUuid?: OrderItem['shippingOrderLineId']) {
+  const orderItemUuidHookParameter = orderItemUuid;
+  const isCreating = useSelector((state: StoreState) =>
```
boliveira, line 35: Since the selector that is being used here is called `isExchangeFilterLoading`, I would change this name to `isLoading`. What do you think? If you agree, do not forget to add this to the description of the commit as it will be a breaking change.

packages/react/src/exchanges/hooks/useExchangeFilters.ts
boliveira: I would also add this: 

`Please notice that if you pass a value for the orderItemUuid property inside exchangeFilterData parameter, the values for isLoading, error and exchangeFilters properties returned by the hook will not be based on it as the hook bases those values on the value of the orderItemUuid that was passed to the hook. If you need to call the create action with a different value for the orderItemUuid property that was passed to the hook, do not rely on the values returned by hook for the isLoading, error and exchangeFilters properties.`

packages/redux/src/exchanges/actions/factories/createExchangeFilterFactory.ts
```diff
@@ -22,15 +28,33 @@ const createExchangeFilterFactory =
   async (
     dispatch: Dispatch<CreateExchangeFilterAction>,
   ): Promise<ExchangeFilter> => {
+    // Although exchangeFilterItems is an array, it only has 1 object at a time, so it's safe to
+    // directly access the first index.
+    const orderItemUuid = data.exchangeFilterItems[0]
```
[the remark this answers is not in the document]
andreecosta: The `orderItemUuid` is never `undefined` in the action. The only type where `orderItemUuid` is undefined, is the new type created for the hook, but even there, you can't call the action without any orderItemUuid (you need to either pass it to the hook or to the function).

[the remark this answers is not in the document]
boliveira: You can if the `exchangeFilterItems` parameter is an empty array.

packages/react/src/exchanges/hooks/useExchangeFilters.ts
```diff
@@ -0,0 +1,111 @@
...
+    orderItemUuidHookParameter
+      ? isExchangeFilterLoading(state, orderItemUuidHookParameter)
+      : false,
+  );
+  const error = useSelector((state: StoreState) =>
+    orderItemUuidHookParameter
+      ? getExchangeFilterError(state, orderItemUuidHookParameter)
+      : null,
+  );
+  const exchangeFilters = useSelector(getExchangeFilters);
```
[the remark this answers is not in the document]
andreecosta: What if I want to pass the `orderItemUuidHookParameter` but still get all the entities? I think it's safer to just return all the entities, and if the user needs they can find a specific one by id on their side.

packages/react/src/exchanges/hooks/useExchange.ts
andreecosta: Done

packages/react/src/exchanges/hooks/useExchangeFilters.ts
andreecosta: Done

packages/react/src/exchanges/hooks/useExchangeFilters.ts
andreecosta: Done

packages/react/src/exchanges/hooks/useExchangeFilters.ts
andreecosta: Done

packages/react/src/exchanges/hooks/useExchangeFilters.ts
andreecosta: Done

packages/react/src/exchanges/hooks/useExchangeFilters.ts
andreecosta: Done

packages/react/src/exchanges/hooks/useExchangeFilters.ts
andreecosta: Fixed and added to the description.

packages/redux/src/exchanges/actions/factories/createExchangeFilterFactory.ts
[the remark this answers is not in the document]
andreecosta: You are right. Done!

packages/react/src/exchanges/hooks/useExchangeFilters.ts
boliveira: This needs to be reexported at the index file. Please, check the file `/Users/bruno.a.oliveira/Code/blackout/packages/react/src/exchanges/hooks/types/useExchangeBookRequest.ts` and make the same changes for this file.

packages/redux/src/exchanges/actions/factories/createExchangeFilterFactory.ts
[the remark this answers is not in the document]
boliveira: Instead of dispatching the error directly, throw the error inside the try/catch block which already addresses that dispatch.

packages/react/src/exchanges/hooks/useExchangeFilters.ts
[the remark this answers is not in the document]
boliveira: Is that a common case? I think that if that is what the user wants, then he can pass undefined to the hook and provide the orderItemUuid value on the parameter to create action. Then we do not have an inconsistency here as other hooks operate in the same way.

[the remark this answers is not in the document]
andreecosta: Done! As talked in private, I've added the option to either return one filter by id or all filters.

boliveira:
Waiting for unit tests to be added for the create action, reducer and getExchangeFilterError/loading when the orderItemUuid parameter is undefined to validate if there is any problem with that.

boliveira (approved):

dnamorim (approved):

diegogriep (approved):

ruifcnunes (approved):

PR #910, opened 2023-11-09
fix: save exchange filters by id

andreecosta:
## Description

BREAKING CHANGE: Exchange filters are now stored under entities by orderItemUuid (shippingOrderLineId). Previously the user wasn't able to fetch multiple filters because each request cleared the previous data, which this PR aims to fix.

## Checklist

## Disclaimer

### DIFF
--- a/packages/core/src/entities/schemas/exchangeFilter.js
+++ b/packages/core/src/entities/schemas/exchangeFilter.js
@@ -1,0 +1,11 @@
+import { schema } from 'normalizr';
+
+const exchangeFiltersSchema = new schema.Entity(
+  'exchangeFilters',
+  {},
+  {
+    idAttribute: value => value.exchangeFilterItems[0].orderItemUuid,
+  },
+);
+
+export default exchangeFiltersSchema;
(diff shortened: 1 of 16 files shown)


jpklzm (approved):

packages/core/src/entities/schemas/exchangeFilter.js
```diff
@@ -0,0 +1,11 @@
+import { schema } from 'normalizr';
+
+const exchangeFiltersSchema = new schema.Entity(
+  'exchangeFilters',
+  {},
+  {
+    idAttribute: value => value.exchangeFilterItems[0].orderItemUuid,
```
pedro-gbf, line 7: In the case of being multiple orderItems being exchanged won't this create a problem since we will only identify the exchangeFilter by the first exchangeFilterItem orderItemUuid?

andreecosta: No, because you can't fetch multiple filters at the same time for more than 1 item. Although the `exchangeFilterItems` is an array, It can only have 1 value at a time.

pedro-gbf: What a marvellous API ❤️, approved

pedro-gbf (approved):

PR #913, opened 2023-11-10
fix: fix `productsList` schema handling of `filterSegments`

### LINKED ISSUE #906: Filter segments on server initial state only looking in the first facet group
## Expected behavior

When there are more than one facet group, when building the filter segments, it should look on all facet groups, not only in the first one. Example: We have 3 categories: Dresses, jumpsuits and knitwear. Each one of these categories has a facetgroup for the sizes, being dresses -> [xs], jumpsuits-> [xs,s,m,l,xl], knitwear -> [s]. 

When building the filter segments only the first array is being used causing for example the s or m size not be found. The code responsible for this is this:

file: `blackout-redux > entities > schemas > productsLists.js`

```if (filterSegment.type === ATTRIBUTES_TYPE) {
        let facetIndex;
        const facetGroup = filteredFacetGroups.find(facet => facet.values[0]?.find(({
          value
        }, i) => {
          if (value === filterSegment.value) {
            facetIndex = i;
            return true;
          }
          return false;
        }));
        const facet = facetIndex ? facetGroup?.values[0]?.[facetIndex] : undefined;
        return {
          ...filterSegment,
          description: filterSegment.description || facet?.description,
          facetId: facet ? getId(facet, facetGroup) : undefined
        };
      }
      const facet = filteredFacetGroups[0]?.values[0]?.find(({
        value
      }) => value === filterSegment.value);
      return {
        ...filterSegment,
        description: filterSegment.description || facet?.description,
        facetId: facet ? getId(facet, filteredFacetGroups?.[0]) : undefined
      };

```
## Actual behavior

To build the filter segments, all the facetgroups should be considered.

There is also a bug here :

`const facet = facetIndex ? facetGroup?.values[0]?.[facetIndex] : undefined;`

If facetIndex is 0 (which means the first position) facet will come as undefined as 0 is a falsy value. The condition should be  `typeof faceIndext !== 'undefined'`

## Steps to reproduce

1. Access [here](https://www-napsandbox2-ft06.blackandwhite-ff.com/en-pt/shopping/clothing?contextfilters=categories:182532&pageindex=1)
2. On the clothing sizes select xs
3. On the clothing sizes select s
4. Check that neither the checkboxes nor the pills are correctly working

boliveira:
## Description

This fixes `productsList` schema handling of the `filterSegments` property which was not finding the correct facet when the filter segment was not pointing to the first facet of the first facet group.

Closes #906 

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/entities/schemas/productsList.ts
+++ b/packages/redux/src/entities/schemas/productsList.ts
@@ -67,7 +67,7 @@ export default new schema.Entity(
           // If the filter segment is an ProductVariantAttribute, filteredFacetGroups will be an array with
           // all attributes so the flow is a little different.
           if (filterSegment.type === ATTRIBUTES_TYPE) {
-            let facetIndex;
+            let facetIndex: number | undefined;
             const facetGroup = filteredFacetGroups.find(facet =>
               facet.values[0]?.find(({ value }, i) => {
                 if (value === filterSegment.value) {
@@ -79,9 +79,10 @@ export default new schema.Entity(
                 return false;
               }),
             );
-            const facet = facetIndex
-              ? facetGroup?.values[0]?.[facetIndex]
-              : undefined;
+            const facet =
+              typeof facetIndex === 'number'
+                ? facetGroup?.values[0]?.[facetIndex]
+                : undefined;
 
             return {
               ...filterSegment,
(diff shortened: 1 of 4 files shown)


diegogriep (approved):

packages/redux/src/entities/schemas/productsList.ts
```diff
@@ -67,7 +67,7 @@ export default new schema.Entity(
           // If the filter segment is an ProductVariantAttribute, filteredFacetGroups will be an array with
           // all attributes so the flow is a little different.
           if (filterSegment.type === ATTRIBUTES_TYPE) {
-            let facetIndex;
+            let facetIndex: undefined | number;
```
dnamorim, line 70: As the facetIndex is expected to be a `number` and not `undefined`, and across the project, we usually use `number | undefined`, can we change this to the suggestion below?

```suggestion
            let facetIndex: number | undefined;
```

dnamorim (changes requested):

packages/redux/src/entities/schemas/productsList.ts
boliveira: Done, thank you!

dnamorim (approved):

PR #930, opened 2023-11-28
feat: add delivery bundles usecheckout

escabora:
## Description

We understand that the responsibility for exposing delivery bundles data lies with `useCheckout`. So I opened this PR in which I get the delivery bundles in the context of redux and expose it in the `useCheckout` hook. I also take the opportunity to add a key to the `useCheckout` that represents which delivery bundle has been selected.

## Checklist

## Disclaimer

### DIFF
--- a/tests/__fixtures__/checkout/checkout.fixtures.mts
+++ b/tests/__fixtures__/checkout/checkout.fixtures.mts
@@ -1054,6 +1054,8 @@ export const deliveryBundlesEntity = {
   },
 };
 
+export const mockDeliveryBundlesEntityDenormalized = [deliveryBundle];
+
 export const deliveryBundleUpgradesEntity = {
   [deliveryBundleId]: {
     [itemId1]: {
@@ -1491,6 +1493,7 @@ export const mockCheckoutState = {
       error: null,
       result: [contextId],
     },
+    deliveryBundles: [deliveryBundleId, '090998'],
   },
   payments: {
     ...mockInitialState.payments,
--- a/tests/__fixtures__/checkout/checkout.fixtures.mts
+++ b/tests/__fixtures__/checkout/checkout.fixtures.mts
@@ -997,7 +997,7 @@ export const checkoutEntity = {
   checkoutOrder: checkoutOrderId,
   id: checkoutId,
   shippingOptions: [shippingOption],
-  deliveryBundles: [deliveryBundleId],
+  deliveryBundles: [deliveryBundleId, '090998'],
   orderStatus: OrderStatusError.NoError,
 };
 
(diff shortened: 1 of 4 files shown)


boliveira (approved):

dnamorim (approved):

tests/__fixtures__/checkout/checkout.fixtures.mts
```diff
@@ -1054,6 +1054,8 @@ export const deliveryBundlesEntity = {
   },
 };
 
+export const mockDeliveryBundlesEntityDenormalized = [deliveryBundle];
```
anapim, line 1057: This still doesn't look okay. Aren't we expecting two delivery bundles in the response?

anapim (changes requested):

tests/__fixtures__/checkout/checkout.fixtures.mts
ivomota: I agree with @anapim. This seems to be not ok. 
Did you tried to add
```js
export const mockDeliveryBundlesEntityDenormalized = [
   deliveryBundle,
   {
    ...deliveryBundle,
    id: '090998',
    name: 'fake bundle',
    isSelected: false,
  }
];
```

I think this should be the result....

anapim: I found the issue. On line 1000, we're missing the second delivery bundle id.

```javascript
export const checkoutEntity = {
  checkoutOrder: checkoutOrderId,
  id: checkoutId,
  shippingOptions: [shippingOption],
  deliveryBundles: [deliveryBundleId, '090998'],
  orderStatus: OrderStatusError.NoError,
};
```

Then @ivomota suggestions should work.

escabora: I made the adjustment. I also found the error in checkoutEntity (line 1000). I needed to add another delivery package id in checkoutEntity. However, when I run the test, I see an error in the redux checkout selectors. I have also solved this problem.

tests/__fixtures__/checkout/checkout.fixtures.mts
```diff
@@ -997,7 +997,7 @@ export const checkoutEntity = {
   checkoutOrder: checkoutOrderId,
   id: checkoutId,
   shippingOptions: [shippingOption],
-  deliveryBundles: [deliveryBundleId],
+  deliveryBundles: [deliveryBundleId, '090998'],
```
dnamorim, line 1000: We are repeating this ID for a second `deliveryBundle` on our mock data several times and in the unit tests 

I would leave the suggestion to have two id's, as we have already one for the `deliveryBundleUpgradeId`:

```
export const deliveryBundleId_1 = '1234';
export const deliveryBundleId_2 = '090998';
```

escabora: Adjustment made!

PR #931, opened 2023-11-28
chore: adding header to support Mock Server

diegogriep:
## Description

The default endpoint will be https://api-next.blackandwhite-ff.com/.

It was created as a header to force requests to go to the origin.

## Checklist

## Disclaimer


packages/client/src/addresses/__fixtures__/getAddressPredictionDetails.fixtures.ts
```diff
@@ -1,7 +1,7 @@
 import { rest, type RestHandler } from 'msw';
 import type { AddressPrediction } from '../types/index.js';
 
-const path = '/api/account/v1/addressesprediction/:predictionId/address';
+const path = '/api-next/account/v1/addressesprediction/:predictionId/address';
```
dnamorim, line 4: To ease the update of these paths in the future, I would suggest defining a constant on configs or for testing configs of the baseURL to use (`/api-next` in this case), and use it like this:

```
import { baseURL } from 'tests/<path-to-localtion>';

const path = '${baseURL}/account/v1/addressesprediction/:predictionId/address';
```

What do you think?

packages/client/src/types/config.types.ts
```diff
@@ -1,3 +1,13 @@
-import type { AxiosRequestConfig } from 'axios';
+import type { AxiosRequestConfig, AxiosRequestHeaders } from 'axios';
 
-export type Config = AxiosRequestConfig & Record<string, unknown>;
+type MockServerProps = {
+  'X-MOCK-SERVER-IGNORE': boolean;
```
ruifcnunes, line 4: Here we need to specify that this header is optional, like this:
```suggestion
  'X-MOCK-SERVER-IGNORE'?: boolean;
```
otherwise if the developer passes in the headers object into a client, it means that it must fully comply with the headers object definition, which will contain this property as required.

diegogriep: Good catch. Thank you!

packages/client/src/types/config.types.ts
```diff
@@ -1,3 +1,13 @@
-import type { AxiosRequestConfig } from 'axios';
+import type { AxiosRequestConfig, AxiosRequestHeaders } from 'axios';
 
-export type Config = AxiosRequestConfig & Record<string, unknown>;
+type MockServerProps = {
+  'X-MOCK-SERVER-IGNORE'?: boolean;
+};
+
+type CustomHeaders = {
```
boliveira, line 7: Export this type as well

packages/client/src/types/config.types.ts
```diff
@@ -1,3 +1,13 @@
...
+  headers?: AxiosRequestHeaders & MockServerProps;
+};
+
+export type Config = AxiosRequestConfig &
```
boliveira, line 11: You will need to Omit the `headers` property from `AxiosRequestConfig` since they will clash here with the type defined on `CustomHeaders` type like this:

```suggestion
export type Config = Omit<AxiosRequestConfig, 'headers'> &
```

packages/client/src/types/config.types.ts
```diff
@@ -1,3 +1,13 @@
-import type { AxiosRequestConfig } from 'axios';
+import type { AxiosRequestConfig, AxiosRequestHeaders } from 'axios';
 
-export type Config = AxiosRequestConfig & Record<string, unknown>;
+type MockServerProps = {
+  'X-MOCK-SERVER-IGNORE'?: boolean;
```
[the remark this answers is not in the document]
diegogriep: We decided this during the refinement session. What are you expecting on the Mock Server side, @ruifcnunes?

packages/client/src/types/config.types.ts
diegogriep: A message like that: `Types of property 'headers' are incompatible`.
Now, I'm using the right types and it was fixed.

dnamorim (approved):

packages/client/src/helpers/client/axiosInstance.ts
```diff
@@ -5,7 +5,7 @@ import axios from 'axios';
  * Create our axios instance to manipulate its defaults.
  */
 export default axios.create({
-  baseURL: '/api',
+  baseURL: '/api-next',
```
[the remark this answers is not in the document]
boliveira: Fixed

packages/client/src/types/config.types.ts
```diff
@@ -1,3 +1,13 @@
-import type { AxiosRequestConfig } from 'axios';
+import type { AxiosRequestConfig, RawAxiosResponseHeaders } from 'axios';
 
-export type Config = AxiosRequestConfig & Record<string, unknown>;
+export type MockServerHeaders = {
+  'X-Mock-Server-Ignore'?: boolean;
+};
+
+export type CustomHeaders = {
```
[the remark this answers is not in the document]
boliveira: Removed this type and defined a new type, `CustomBlackoutClientAxiosRequestConfig` which will now take care of defining the headers. If we need in the future to have that type exported, we can reintroduce it.

ruifcnunes (approved):

PR #933, opened 2023-11-29
chore(react-client): introduces new types to checkout order / session

ghost:
## Description

This PR adds new types for the checkout order and checkout session. These types will be necessary for a new checkout feature: new impersonation flow to post an order

## Checklist

## Disclaimer


packages/client/src/checkout/types/checkoutOrder.types.ts
```diff
@@ -78,4 +78,7 @@ export type CheckoutOrder = {
   shippingMode: ShippingMode;
   paymentIntentId?: PaymentIntent['id'];
   promotionEvaluationId?: PromotionEvaluationId;
+  customerId?: number;
```
anapim, line 81: I don't think these types belong here. From what I understood they're used on the request to create the order and not the response.

packages/client/src/checkout/types/checkoutSession.types.ts
```diff
@@ -53,4 +53,6 @@ export type CheckoutSession = {
   promotionEvaluationId?: string;
   totalCredit?: number;
   promocode?: string;
+  customerId?: number;
```
anapim, line 56: Same here.

anapim (changes requested):

packages/client/src/checkout/types/checkoutOrder.types.ts
hichico: Changed to the post request. Let me know if this is what was intended.

anapim: Now it looks ok!
But on further investigation I think I've missed something before. `customerId` should be part of both the `checkoutOrder` and `checkoutSession` (as you had previously but **only** for `customerId`)

packages/client/src/checkout/types/postCheckoutOrder.types.ts
```diff
@@ -18,6 +18,9 @@ export type PostCheckoutOrderItem = {
   quantity: number;
   customAttributes: string;
   productAggregatorId: number;
+  tenantId?: number;
+  customerId?: number;
+  isGuestUser?: boolean;
```
anapim, line 23: Move to `PostCheckoutOrderdata`

anapim (changes requested):

anapim (approved):

packages/client/src/checkout/types/postCheckoutOrder.types.ts
```diff
@@ -9,6 +9,9 @@ export type PostCheckoutOrderData = {
   guestUserEmail?: string;
   shippingMode?: ShippingMode;
   metadata?: PostCheckoutOrderMetadata;
+  tenantId?: number;
+  customerId?: number;
+  isGuestUser?: boolean;
```
boliveira, line 14: Question: Does the server validate if invalid combination of parameters are sent? For example, I could send `guestUserEmail` with a value but set `isGuestUser` to false. What will happen in that case? If it is invalid, then I think this type should be divided in 2: One when `guestUserEmail` is a string which will restrict the type of `isGuestUser` to `true | undefined` and another where `guestUserEmail` is undefined and `isGuestUser` can be then `boolean | undefined`.

boliveira, line 14: Question: Does the server validate if invalid combination of parameters are sent? For example, I could send `guestUserEmail` with a value but set `isGuestUser` to false. What will happen in that case? If it is invalid, then I think this type should be divided in 2: One when `guestUserEmail` is a string which will restrict the type of `isGuestUser` to `true | undefined` and another where `guestUserEmail` is undefined and `isGuestUser` can be then `boolean | undefined`.

packages/client/src/checkout/types/postCheckoutSession.types.ts
```diff
@@ -11,6 +11,9 @@ export type PostCheckoutSessionData = {
   metadata?: CheckoutSessionMetadata;
   successUrl: string;
   cancelUrl: string;
+  tenantId?: number;
+  customerId?: number;
+  isGuestUser?: boolean;
```
boliveira, line 16: Same as https://github.com/Farfetch/blackout/pull/933/files#r1413557582

boliveira, line 16: Same as https://github.com/Farfetch/blackout/pull/933/files#r1413557582

dnamorim:
This PR is intended to go to `next` (prerelease) or to `main` (client v2 / redux v1)? 
Because the target branch is set as `next`, but the milestone is `dev-main-28-11-2023`

dnamorim:
This PR is to be released on `next` (prerelease) or on `main` (v2 client / v1 redux)?

Because the target branch is set as `next`, but the milestone is set as `dev-main-28-11-2023`

ivomota (approved):

hichico:
> This PR is to be released on `next` (prerelease) or on `main` (v2 client / v1 redux)?
> 
> Because the target branch is set as `next`, but the milestone is set as `dev-main-28-11-2023`

Hi @dnamorim 👋 
This branch is aiming for the next branch so we can ditch the appointed milestone.

packages/client/src/checkout/types/postCheckoutOrder.types.ts
hichico: Waiting for response from the back-end team. Will answer you as soon as I have the right info.

PR #949, opened 2023-12-22
feat: add support to consent management protocol v2 to `main`

danielbento92:
## Description

- add support to consent management protocol v2 (analytics)

### Dependencies

None.

## Checklist

## Disclaimer

### DIFF
--- a/packages/react/src/__tests__/__snapshots__/index.test.ts.snap
+++ b/packages/react/src/__tests__/__snapshots__/index.test.ts.snap
@@ -248,6 +248,10 @@ Object {
     "GA": [Function],
     "GA4": [Function],
     "GTM": [Function],
+    "GoogleConsentType": Object {
+      "Denied": "denied",
+      "Granted": "granted",
+    },
     "Integration": [Function],
     "Omnitracking": [Function],
     "Riskified": [Function],
--- a/packages/react/src/analytics/integrations/GA4/GA4.ts
+++ b/packages/react/src/analytics/integrations/GA4/GA4.ts
@@ -15,6 +15,7 @@
 import {
   PageType as analyticsPageTypes,
   TrackType as analyticsTrackTypes,
+  type ConsentData,
   type EventData,
   integrations,
   type LoadIntegrationEventData,
@@ -36,6 +37,7 @@ import {
   OPTION_DATA_LAYER_NAME,
   OPTION_DEBUG_MODE,
   OPTION_ENABLE_AUTOMATIC_PAGE_VIEWS,
+  OPTION_GOOGLE_CONSENT_CONFIG,
   OPTION_LOAD_SCRIPT_FUNCTION,
   OPTION_MEASUREMENT_ID,
   OPTION_NON_INTERACTION_EVENTS,
@@ -46,6 +48,7 @@ import {
   OPTION_SET_CUSTOM_USER_ID_PROPERTY,
 } from './constants.js';
 import { each, get, merge } from 'lodash-es';
+import { GoogleConsentMode } from '../shared/index.js';
 import { validateFields } from './validation/optionsValidator.js';
 import defaultSchemaEventsMap from '../shared/validation/eventSchemas.js';
 import eventValidator from '../shared/validation/eventValidator.js';
@@ -70,6 +73,7 @@ import type {
  * Google Analytics 4 Integration.
  */
 class GA4 extends integrations.Integration<GA4IntegrationOptions> {
+  private googleConsentMode!: GoogleConsentMode;
   private enableAutomaticPageViews!: boolean;
   private schemaEventsMap!: Schemas;
   private initializePromise!: Promise<void>;
@@ -100,7 +104,7 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
     strippedDownAnalytics: StrippedDownAnalytics,
   ) {
     super(options, loadData, strippedDownAnalytics);
-    this.initialize(options);
+    this.initialize(options, loadData);
     this.onSetUser(loadData);
   }
 
@@ -111,9 +115,13 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
    * Initializes member variables from options and tries to initialize Google
    * Analytics 4.
    *
-   * @param options - Options passed for the GA4 integration.
+   * @param options   - Options passed for the GA4 integration.
+   * @param loadData              - Analytics's load event data.
    */
-  initialize(options: GA4IntegrationOptions) {
+  initialize(
+    options: GA4IntegrationOptions,
+    loadData: LoadIntegrationEventData,
+  ) {
     this.optionsValidationResultsMap = validateFields(options);
 
     this.measurementId = options[OPTION_MEASUREMENT_ID];
@@ -148,7 +156,31 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
       true,
     );
 
+    this.googleConsentMode = new GoogleConsentMode(
+      this.getDataLayerName(),
+      loadData.consent,
+      get(options, OPTION_GOOGLE_CONSENT_CONFIG),
+    );
+
     this.loadGtagScript(options);
+  }
+
+  getDataLayerName(): string {
+    return get(this.options, OPTION_DATA_LAYER_NAME, DEFAULT_DATA_LAYER_NAME);
+  }
+
+  /**
+   * Sets the consent object. This method is called by analytics whenever the consent
+   * changes, so there's no need to validate if it has changed or not.
+   *
+   * @param consent - Object to be written on the dataLayer.
+   *
+   * @returns This allows chaining of class methods.
+   */
+  override setConsent(this: this, consent: ConsentData): this {
+    this.googleConsentMode.updateConsent(consent);
+
+    return this;
   }
 
   /**
@@ -725,9 +757,7 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
    * @param options - User configured options.
    */
   internalLoadScript(options: GA4IntegrationOptions) {
-    const customDataLayerAttr = options[OPTION_DATA_LAYER_NAME]
-      ? options[OPTION_DATA_LAYER_NAME]
-      : DEFAULT_DATA_LAYER_NAME;
+    const customDataLayerAttr = this.getDataLayerName();
     const debugMode = options[OPTION_DEBUG_MODE] || false;
     const script = document.createElement('script');
 
--- a/packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
+++ b/packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
@@ -1,0 +1,159 @@
+import { type ConsentData } from '@farfetch/blackout-analytics';
+import {
+  type GoogleConsentCategoryConfig,
+  type GoogleConsentModeConfig,
+  GoogleConsentType,
+} from './types.js';
+import { isEqual, omit } from 'lodash-es';
+
+/**
+ * GoogleConsentMode handles with Google Consent Mode v2.
+ */
+export class GoogleConsentMode {
+  private dataLayer!: string; // Stores different data layer names
+  private config?: GoogleConsentModeConfig; // Stores default or customized consent category mappings
+  private configExcludingRegionsAndWaitForUpdate!: Record<
+    string,
+    GoogleConsentCategoryConfig
+  >; // exclude not consent properties from config
+  private lastConsent?: Record<
+    string,
+    Array<string> | string | number | undefined
+  >;
+
+  constructor(
+    dataLayer: string,
+    initConsent: ConsentData | null,
+    config?: GoogleConsentModeConfig,
+  ) {
+    this.dataLayer = dataLayer;
+    this.config = config;
+
+    // select only the Google Consent Elements
+    this.configExcludingRegionsAndWaitForUpdate = omit(this.config || {}, [
+      'waitForUpdate',
+      'regions',
+    ]);
+
+    this.loadDefaults(initConsent);
+  }
+  /**
+   * Write google consent default values to dataLayer.
+   *
+   * @param initConsent - The init consent data to be set.
+   */
+  private loadDefaults(initConsent: ConsentData | null) {
+    if (this.config) {
+      const initialValue: Record<string, string | number> = {};
+
+      if (this.config.waitForUpdate) {
+        initialValue['wait_for_update'] = this.config.waitForUpdate;
+      }
+
+      // Obtain default google consent registry
+      const consentRegistry = Object.keys(
+        this.configExcludingRegionsAndWaitForUpdate,
+      ).reduce(
+        (result, consentKey) => ({
+          ...result,
+          [consentKey]:
+            this.configExcludingRegionsAndWaitForUpdate[consentKey]?.default ||
+            GoogleConsentType.Denied,
+        }),
+        initialValue,
+      );
+
+      // Write default consent to data layer
+      this.write('consent', 'default', consentRegistry);
+
+      // write regions to data layer if they exists
+      this.config.regions?.forEach(region => {
+        this.write('consent', 'default', region);
+      });
+
+      this.updateConsent(initConsent);
+    }
+  }
+
+  /**
+   * Update consent.
+   *
+   * @param consentData - The consent data to be set.
+   */
+  updateConsent(consentData: ConsentData | null) {
+    if (this.config) {
+      // Dealing with null or undefined consent values
+      const safeConsent = consentData || {};
+
+      // Fill consent value into consent element, using analytics consent categories
+      const consentRegistry = Object.keys(
+        this.configExcludingRegionsAndWaitForUpdate,
+      ).reduce((result, consentKey) => {
+        let consentValue = GoogleConsentType.Denied;
+        const consent = this.configExcludingRegionsAndWaitForUpdate[consentKey];
+
+        if (consent) {
+          // has consent config key
+
+          if (consent.getConsentValue) {
+            // give priority to custom function
+            consentValue = consent.getConsentValue(safeConsent);
+          } else if (
+            consent?.categories !== undefined &&
+            consent.categories.every(consent => safeConsent[consent])
+          ) {
+            // The second option to assign value is by categories list
+            consentValue = GoogleConsentType.Granted;
+          }
+        }
+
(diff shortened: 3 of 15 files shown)


packages/react/src/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -248,9 +248,14 @@ Object {
     "GA": [Function],
     "GA4": [Function],
     "GTM": [Function],
+    "GoogleConsentMode": [Function],
```
boliveira, line 251: Is this supposed to be available to users? From what I can understand, this is only used by some integrations internally.

packages/react/src/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -248,9 +248,14 @@ Object {
...
     "Integration": [Function],
     "Omnitracking": [Function],
     "Riskified": [Function],
+    "googleConsentMode": GoogleConsentMode {
```
boliveira, line 255: Same here, this shouldn't be exposed to end-users.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/index.ts
```diff
@@ -0,0 +1,5 @@
+export {
+  GoogleConsentMode,
+  default as googleConsentMode,
```
boliveira, line 3: I think we can improve these namings as they are too similar to understand what each one represents.

packages/react/src/analytics/integrations/GA4/GA4.ts
```diff
@@ -148,9 +150,36 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
       true,
     );
 
+    const dataLayer = get(
```
boliveira, line 153: Since this is being used in `setConsent` as well, what do you think of creating a helper function called something like `resolveDataLayerName` which receives the options as an argument and returns the datalayer name which will be used?

packages/react/src/analytics/integrations/GA4/GA4.ts
```diff
@@ -148,9 +150,36 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
...
+      this.options,
+      OPTION_DATA_LAYER_NAME,
+      DEFAULT_DATA_LAYER_NAME,
+    );
+
+    googleConsentMode.addIntegration(this.loadData.consent, dataLayer);
```
boliveira, line 159: I do not understand why this function is called `addIntegration` since we are adding `dataLayer`s here. Also, I think the order of the arguments should be reversed, i.e., dataLayer should be first.

packages/react/src/analytics/integrations/GA4/GA4.ts
```diff
@@ -148,9 +150,36 @@ class GA4 extends integrations.Integration<GA4IntegrationOptions> {
...
+   * @returns This allows chaining of class methods.
+   */
+  override setConsent(this: this, consent: ConsentData): this {
+    const dataLayer = get(
+      this.options,
+      OPTION_DATA_LAYER_NAME,
+      DEFAULT_DATA_LAYER_NAME,
+    );
+
+    googleConsentMode.setConsents(consent, dataLayer);
```
boliveira, line 178: Why no call `setConsentForDataLayer`? Also, the arguments should be reversed, i.e., dataLayer should be first.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,124 @@
...
+  private gtag(
+    dataLayer: string,
+    consent: ConsentData | null,
+    command = 'default',
+  ) {
+    if (this.enabled && typeof window !== 'undefined') {
+      // @ts-ignore
+      window[dataLayer] = window[dataLayer] || [];
+
+      function internalGtag() {
```
boliveira, line 110: Why is it necessary to have this function? I think this will complicate the code a little bit and the performance will be affected since this function will always be recreated when the gtag method is called.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * Handles consent by updating the data layer with consent information.
+   *
+   * @param dataLayer - The name of the data layer.
+   * @param consent - The consent data.
+   * @param command - Optional. The command for consent handling.
+   */
+  private gtag(
```
boliveira, line 101: I am not sure if this is the best name for this function, I would be more specific and set a name corresponding more to what the function is supposed to do.

packages/react/src/__tests__/__snapshots__/index.test.ts.snap
danielbento92: Users can need to interact with consent mode, for example to define which consents would be mapped to their categories or disable GoogleConsentMode module itself.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/index.ts
danielbento92: I follow the same pattern with Analytics/AnalyticsWeb (instances) namings, but I understand that it might be difficult to distinguish. Do you have any suggestions?

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,124 @@
...
+   */
+  private getConsentValues(consents: ConsentData) {
+    // Dealing with null or undefined consent values
+    const consentList = consents || {};
+
+    if (this.consentCategoriesMapping) {
+      return Object.keys(this.consentCategoriesMapping).reduce(
+        (result, consentKey) => ({
+          ...result,
+          [consentKey]: this.consentCategoriesMapping?.[consentKey]?.every(
```
boliveira, line 61:
```suggestion
          [consentKey]: this.consentCategoriesMapping!.[consentKey]?.every(
```

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * Retrieves consent values for each category based on provided consents.
+   *
+   * @param consents - The consent data.
+   *
+   * @returns Consent values for each category.
+   */
+  private getConsentValues(consents: ConsentData) {
```
boliveira, line 53: Use singular here please, since it is the same data type that is being used in setConsent throughout the package.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,124 @@
...
+    // Dealing with null or undefined consent values
+    const consentList = consents || {};
```
boliveira, line 55:
```suggestion
    const safeConsent = consents || {};
```

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+  }
+
+  /**
+   * Sets consents for specified data layers.
+   *
+   * @param consents - The consent data to be set.
+   * @param dataLayer - The specific data layer where consents should be set.
+   */
+  setConsents(consents: ConsentData, dataLayer?: string) {
```
boliveira, line 86: setConsent is more appropriate as this is the same idea that is being used in `integration.setConsent`.

packages/react/src/analytics/integrations/GA4/GA4.ts
danielbento92: SetConsent method, work in two ways: with or without datalayer property.

If datalayer property set, then googleConsentMode should sent only to this datalayer. If property is empty, the method dispatch for each datalayer added do google Consent Mode.


For example, If I want to send manually consents to gtm and google analytics, I could by send this parameter undefined.

packages/react/src/analytics/integrations/GA4/GA4.ts
```diff
@@ -72,6 +73,7 @@ import type {
  * Google Analytics 4 Integration.
  */
 class GA4 extends integrations.Integration<GA4IntegrationOptions> {
+  protected googleConsentMode?: GoogleConsentMode;
```
boliveira, line 76: You can use the ! operator to tell typescript that this will be set after the constructor is called.

```suggestion
  protected googleConsentMode!: GoogleConsentMode;
```

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -5,42 +5,18 @@ import { type GoogleConsentModeMappings } from './types.js';
  * GoogleConsentMode handles with Google Consent Mode v2.
  */
 export class GoogleConsentMode {
-  private dataLayers!: Array<string>; // Stores different data layer names
-  private enabled!: boolean; // enable or disable CMP integration
+  private dataLayers!: string; // Stores different data layer names
```
boliveira, line 8: Since this is a single datalayer now, you can rename this to singular:

```suggestion
  private dataLayer!: string; // Stores different data layer names
```

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -50,7 +26,7 @@ export class GoogleConsentMode {
    *
    * @returns Consent values for each category.
    */
-  private getConsentValues(consents: ConsentData) {
+  private getConsentValues(consents: ConsentData | null) {
```
boliveira, line 29: Treat consent as singular as it is like that everywhere else

```suggestion
  private getConsentValues(consents: ConsentData | null) {
```

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -5,42 +5,18 @@ import { type GoogleConsentModeMappings } from './types.js';
...
-   *
-   * @param consentMapping - Custom mappings to be set.
-   */
-  setConsentCategoriesMapping(consentMapping: GoogleConsentModeMappings) {
-    this.consentCategoriesMapping = consentMapping;
-  }
+  constructor(
+    dataLayer: string,
+    consent: ConsentData | null,
+    consentCategoriesMapping?: GoogleConsentModeMappings,
```
boliveira, line 14:
```suggestion
    consentCategoriesMappings?: GoogleConsentModeMappings,
```

boliveira, line 14:
```suggestion
    consentCategoriesMappings?: GoogleConsentModeMappings,
```

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -98,27 +69,23 @@ export class GoogleConsentMode {
    * @param consent - The consent data.
    * @param command - Optional. The command for consent handling.
    */
-  private gtag(
+  private write(
     dataLayer: string,
```
boliveira, line 73: This parameter is unnecessary now since you can access from the this instance.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/__tests__/GoogleConsentMode.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      mockConsent,
+      consentMappings,
+    );
+
+    expect(googleConsent).toBeInstanceOf(GoogleConsentMode);
+
+    expect(window.dataLayer).toMatchSnapshot();
+  });
+
+  it('should initialize GoogleConsentMode without consents and grant consents before', () => {
```
boliveira, line 37: Is this description correct? It seems it is missing something.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/types.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+export interface GoogleConsentOptions {
+  ad_storage: Array<string>;
+  ad_user_data: Array<string>;
+  ad_personalization: Array<string>;
+  analytics_storage: Array<string>;
+}
+
+export interface GoogleConsentMappings {
```
boliveira, line 12: This is the same as the previous type

boliveira: Also, this type does not need to be exported and I would give it another name that is more distinguishable like `GoogleConsentMappingsBase`

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * @param consent - The consent data to be set.
+   */
+  updateConsent(consent: ConsentData) {
+    this.write('consent', 'update', this.getConsentValues(consent));
+  }
+
+  /**
+   * Handles consent by updating the data layer with consent information.
+   *
```
boliveira, line 69: Add a commentary saying why the arguments are being ignored so other developers will know what is the reason behind it.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/__tests__/GoogleConsentMode.test.ts
danielbento92: Updated description to: Initialize GoogleConsentMode without any consents. Later, set all consents and expect that all results are granted.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/types.ts
danielbento92: Fixed

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
danielbento92: Fixed

boliveira (approved):

talbertosilva (approved):

glaucioripol (approved):

dnamorim (approved):

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,154 @@
...
+  constructor(
+    dataLayer: string,
+    initConsent: ConsentData | null,
+    config?: GoogleConsentModeConfig,
+  ) {
+    this.dataLayer = dataLayer;
+    this.config = config;
+
+    // select only the Google Consent Elements
+    this.consentConfig = omit(this.config || {}, [
```
boliveira, line 30: I think this name is misleading since there is a `this.config` variable that is of type GoogleConsentModeConfig. What about renaming this to configExcludingRegionsAndWaitForUpdate? It is a long name but more distinguishable.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
```diff
@@ -0,0 +1,154 @@
...
+   * @param consentParams - The consent arguments.
+   */
+  private write(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    consentCommand: 'consent',
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    command: 'default' | 'update',
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    consentParams:
+      | Record<string, Array<string> | string | number | undefined>
```
boliveira, line 130: Is this type correct? I think this is a Record<string, GoogleConsentType> type.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/types.ts
```diff
@@ -0,0 +1,34 @@
+import { type ConsentData } from '@farfetch/blackout-analytics';
+
+export enum GoogleConsentType {
+  Granted = 'granted',
+  Denied = 'denied',
+}
+
+export type GoogleConsentRegionConfig = {
```
boliveira, line 8: This type is not being used

```suggestion
export type GoogleConsentRegionConfig = {
```

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
danielbento92: We need to allow the extra possibilities like region or wait_for_update, which is an Array of strings and number.

packages/react/src/analytics/integrations/shared/GoogleConsentMode/types.ts
danielbento92: fixed

packages/react/src/analytics/integrations/shared/GoogleConsentMode/types.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export type GoogleConsentMappingsBase<T> = {
+  ad_storage: T;
+  ad_user_data: T;
+  ad_personalization: T;
+  analytics_storage: T;
+};
+
+export type RegionGoogleConsentMappings = Partial<
```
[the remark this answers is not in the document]
danielbento92: fixed

packages/react/src/analytics/integrations/shared/GoogleConsentMode/GoogleConsentMode.ts
danielbento92: I think this two options excluded(omitted), was the only one properties not "consent" values.
Could change for example:

export type GoogleConsentModeConfig =
   {
   consent: GoogleConsentMappingsBase<GoogleConsentCategoryConfig>
    regions?: Array<GoogleConsentRegionConfig>;
    wait_for_update?: number;
  };

PR #957, opened 2024-01-05
fix(react): add product aggregator id to add item to bag

danpadrinom:
## Description

This adds the `productAgreggatorId` as an optional parameter to the `addItem` in the `useBag` hook.

### Dependencies

None.

## Checklist

## Disclaimer


packages/react/src/bags/hooks/useBag.ts
```diff
@@ -300,6 +300,10 @@ const useBag = (options: UseBagOptions = {}) => {
     (
       productId: Product['result']['id'],
       { quantity, sizeId }: { quantity: number; sizeId: number },
+      productAggregatorId?: Exclude<
+        BagItemDenormalized['productAggregator'],
+        null
+      >['id'],
```
dnamorim, line 306: Adding the `productAggregatorId` as a parameter will force all the functions that use the `metadata` and/or `config` parameters to change to add `undefined` on their calls.

If the second parameter of `addItem` accepted the new optional parameter `productAggregatorId`, this new change would not force anyone to change the addItem callback:

```tsx
const addItem = useCallback(
    (
      productId: Product['result']['id'],
      {
        quantity,
        sizeId,
        productAggregatorId,
      }: {
        quantity: number;
        sizeId: number;
        productAggregatorId:
          | Exclude<BagItemDenormalized['productAggregator'], null>['id']
          | undefined;
      },
      metadata?: BagItemActionMetadata,
      config: Config | undefined = fetchConfig,
    )
// ...
```

boliveira: I agree with @dnamorim and I would also add that the types of productId, quantity, sizeId, productAggregatorId and metadata can be derived from the `PostBagItemData` type as they are ultimately based on.

packages/react/src/bags/hooks/__tests__/useBag.test.tsx
```diff
@@ -617,7 +646,7 @@ describe('useBag', () => {
 
         await addItem(
           mockProductId,
-          { quantity: 1, sizeId: 1 },
+          { quantity: 1, sizeId: 1, productAggregatorId: undefined },
```
dnamorim, line 649: In here, I would suggest to omit the `productAggregatorId` to ensure that we don't provide this property, by default it will assume the `undefined` value

packages/react/src/bags/hooks/useBag.ts
danpadrinom: Great catch! Thanks, updated 👍

packages/react/src/bags/hooks/__tests__/useBag.test.tsx
danpadrinom: Agree, I've updated similar cases as well!

boliveira (approved):

packages/react/src/bags/hooks/__tests__/useBag.test.tsx
```diff
@@ -629,7 +654,6 @@ describe('useBag', () => {
             authCode: undefined,
             customAttributes: '',
             merchantId: mockMerchantId,
-            productAggregatorId: undefined,
```
dnamorim, line 632: You can keep this line to ensure that the productAggregatorId when we do not specify it keeps being undefined.

dnamorim (approved):

PR #960, opened 2024-01-09
fix: commerce pages in ssr

richard190m:
## Description

Fix commerce pages on server side by:
* Fix serverInitialState to include strategy prop
* Fix hooks useCommercePages and useContentPages to use slug as hash to redux store  

## Checklist

## Disclaimer


joaoprcosta (approved):

vinicaval (approved):

edupaulos (approved):

packages/redux/src/contents/actions/factories/fetchCommercePagesFactory.ts
```diff
@@ -39,17 +39,19 @@ const fetchCommercePagesFactory =
     let hash: string | undefined;
 
     try {
+      const { slug, ...queryWithoutSlug } = query;
+
       hash = generateContentHash({
         contentTypeCode: ContentTypeCode.CommercePages,
-        ...query,
+        codes: slug,
```
boliveira, line 46: This will make requests for commerce pages containing different query parameters to be indexed on the same entry in the redux store. Is this ok?

packages/redux/src/contents/actions/factories/fetchCommercePagesFactory.ts
```diff
@@ -29,7 +29,7 @@ import type { Dispatch } from 'redux';
 const fetchCommercePagesFactory =
   (getCommercePages: GetCommercePages) =>
   (
-    query: QueryCommercePages,
+    query: QueryCommercePagesWithSlug,
```
boliveira, line 32: This will be a breaking change, can't you calculate a default value for the slug or at least make it optional?

packages/react/src/contents/hooks/types/useCommercePages.types.ts
```diff
@@ -2,6 +2,7 @@ import type { CommercePagesRankingStrategy } from '@farfetch/blackout-redux';
 import type { Config, QueryCommercePages } from '@farfetch/blackout-client';
 
 export interface UseCommercePagesOptions extends QueryCommercePages {
+  slug: string;
```
boliveira, line 5: This will be a breaking change as well, can't it be at least defined as optional?

packages/redux/src/contents/serverInitialState.ts
```diff
@@ -15,28 +18,36 @@ import type { ServerInitialState } from '../types/serverInitialState.types.js';
...
  */
-const serverInitialState: ServerInitialState = ({ model }) => {
+const serverInitialState: ServerInitialState = ({ model, strategy }) => {
   if (!get(model, 'searchContentRequests')) {
     return { contents: INITIAL_STATE_CONTENT };
   }
 
   const { searchContentRequests, slug, seoMetadata, subfolder } = model;
+  const url = subfolder !== '/' ? slug?.replace(subfolder, '') : slug;
+  const normalizedUrl = url
```
boliveira, line 28: Is this necessary? If json=true is passed, the response returned by FO will not be html, so the application will not be run in this case.

packages/redux/src/contents/serverInitialState.ts
```diff
@@ -15,28 +18,36 @@ import type { ServerInitialState } from '../types/serverInitialState.types.js';
...
+    const code = isCommercePage ? normalizedUrl : codes?.[0];
     const hash = generateContentHash({
-      codes: firstSearchResponseItem.code,
-      contentTypeCode: firstSearchResponseItem.contentTypeCode,
+      codes: code,
+      contentTypeCode: contentTypeCode,
     });
 
+    if (isCommercePage) {
+      response = applyCommercePagesRankingStrategy(searchResponse, strategy);
```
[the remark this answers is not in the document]
boliveira: Ignore this please, I mistakenly confused `.entries` with `Object.entries()` call. But testing in a ClearScript environment is still a good idea since this code might perform differently when running there.

packages/redux/src/contents/actions/factories/fetchCommercePagesFactory.ts
richard190m: This only generate an hash differently. The hash for commerce pages need to be the url instead of query params

packages/react/src/contents/hooks/useContentPage.ts
```diff
@@ -24,14 +24,20 @@ const useContentPage = <T = [ComponentType]>(
...
+    enableAutoFetch = true,
+    fetchConfig,
+    isCommercePage = false,
+  } = options;
 
   const query = useMemo(
     () => ({
-      contentTypeCode: ContentTypeCode.ContentPage,
-      codes: fetchQuery.slug.split('?')[0] as string,
+      contentTypeCode: isCommercePage
```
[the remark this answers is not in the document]
richard190m: Currently, to distinguish from the fetch to commerce pages endpoint directly we use the contentType commercePages and for the request to Serverless API, we use contentType contentPages, to generate the hash

packages/react/src/contents/hooks/types/useCommercePages.types.ts
richard190m: As mentioned before, this is only used by BG and this will fix the commerce pages in this package version

packages/react/src/contents/hooks/useContentPage.ts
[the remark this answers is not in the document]
boliveira: Ok, but in that case, why not use the `useCommercePages` hook by itself? I think that this signature is leaking implementation details that the consumer should not have to worry about. In my opinion, the correct solution would be for the `serverInitialState` to generate entries for both contentTypes in the store so the consumer does not need to worry about this.
What I mean is, the serverInitialState would set an entry for the `commercePages` and `contentPages` contentTypes containing the same data. I think the have the same data format so it would work. If that assumption is not correct, then we should adapt the responses from FO to the proper format.

PR #970, opened 2024-01-17
fix: fix recentlyViewedProducts reducer to sort descending

boliveira:
## Description

This fixes recentlyViewedProducts reducer to sort the products by descending order instead of ascending.

## Checklist

## Disclaimer

### DIFF
--- a/packages/redux/src/products/reducer/recentlyViewedProducts.ts
+++ b/packages/redux/src/products/reducer/recentlyViewedProducts.ts
@@ -1,7 +1,7 @@
 import * as actionTypes from '../actionTypes/index.js';
 import * as authenticationActionTypes from '../../users/authentication/actionTypes.js';
 import { type AnyAction, combineReducers, type Reducer } from 'redux';
-import { omit, sortBy, uniqBy } from 'lodash-es';
+import { omit, orderBy, uniqBy } from 'lodash-es';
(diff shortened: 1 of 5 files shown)


dnamorim (approved):

packages/redux/src/products/reducer/recentlyViewedProducts.ts
```diff
@@ -55,9 +55,10 @@ const result = (
         remote: action.payload,
         pagination: omit(action.payload, 'entries') as RecentlyViewedProducts,
         computed: uniqBy(
-          sortBy(
+          orderBy(
             [...action.payload.entries, ...computed],
             entry => new Date(entry.lastVisitDate),
+            'desc',
```
diegogriep, line 61: A suggestion will be the order to be a param on `result()` arguments.
i.e.:` result(state, action, order = 'desc')`

diegogriep (approved):

packages/redux/src/products/reducer/recentlyViewedProducts.ts
boliveira: The reducer is invoked by redux directly, we cannot pass other arguments to it directly.
We could have this in the `getRecentlyViewedProducts` selector but I think the descending order is what makes sense for all cases. If there is a need to have it ordered differently, the consumer will have to do it.

PR #975, opened 2024-01-24
feat: add parity between ga4 and gtm for `v1`

talbertosilva:
## Description

- Added changes to guarantee a better parity between ga4 and gtm tracking, this involves the addition of events that weren't being tracked on gtm and the respective parameters based on ga4 trackings.
- Changed two utils functions that are directly related with the trackings as well (getProductData and getProductCategory)
- Implemented and adjusted the unit tests affected by this changes.

### Dependencies

None.

## Checklist

## Disclaimer


packages/react/src/analytics/integrations/GTM/eventsMapper.js
```diff
@@ -5,63 +5,326 @@
...
+const getEventTotalValue = (eventProperties, items) => {
+  // There could be cases where the client is not using the bag middleware and wants to pass a value.
+  if (typeof eventProperties.value === 'number') {
+    return eventProperties.value;
+  }
+
+  const value = items?.reduce((acc, item) => {
+    const lineValue =
+      (get(item, 'price', 0) - get(item, 'discount', 0)) *
+      get(item, 'quantity', 1);
```
glaucioripol, line 30: why are you using the get here, just asking to learn?

What do you think about using an Object with these values? 
like it:

```typescript

const { price, discount, quantity } = item ?? {
  price: 0,
  discount: 0,
  quantity: 1,
};

const lineValue = (price - discount) * quantity;

```

it is closest to the vanilla JS, avoiding using a library.

packages/react/src/analytics/integrations/GTM/eventsMapper.js
```diff
@@ -5,63 +5,326 @@
...
       list: properties.list,
       currency: properties.currency,
+      from: properties.from,
+      error: properties.error,
+      itemListId: properties.listId,
+      itemListName: properties.list,
+      items: getProductItemsFromEvent(properties),
+      filters: properties.filters
+        ? JSON.stringify(properties.filters)
+        : undefined,
```
glaucioripol, line 270: When the `JSON.stringify` is used in an undefined value it will return an undefined, so what do you think about using `JSON.stringify` directly in this line ?
`filters: JSON.stringify(properties.filters)`

Example of behavior:

Screenshot 2024-01-24 at 10 32 01

packages/react/src/analytics/integrations/GTM/utils.js
```diff
@@ -54,13 +55,24 @@ export const getContextParameters = context => ({
...
  */
 export const getProductCategory = categories => {
-  if (isArray(categories)) {
-    return categories;
-  }
+  let productCategories = isArray(categories)
+    ? [...categories]
+    : isString(categories)
+    ? categories.split('/')
+    : [];
```
glaucioripol, line 62: What do you think about spreading it in more lines and variables to improve the readability?

Because nested ternary makes the code more complex to read. 
Does it make sense to you?

packages/react/src/analytics/integrations/GTM/eventsMapper.js
boliveira: That will only work if item is undefined/null but we need to account for the cases where the item is set but some of the properties are not set (price, discount and quantity).

packages/react/src/analytics/integrations/GTM/eventsMapper.js
boliveira: If properties.filters is `null`, JSON.stringify will return 'null' which we do not want.

dnamorim (approved):

packages/react/src/analytics/integrations/GTM/utils.js
talbertosilva: Makes sense! Fixed 👍🏼

packages/react/src/analytics/integrations/GTM/eventsMapper.js
```diff
@@ -72,79 +337,144 @@ const eventsMapper = {
...
     };
   },
-  [eventTypes.CHECKOUT_STARTED]: data => {
+  [eventTypes.PLACE_ORDER_FAILED]: data => {
     const properties = getEventProperties(data);
 
     return {
       products: get(properties, 'products', []).map(getProductData),
-      orderId: properties.orderId,
-      total: properties.total,
```
danielbento92, line 81: I think this property `total` could be missing.
I understand that was "replaced" by value, but for full compatibility, I think we need to guarantee that `total` stay on payload, so I suggest on new method, add `total` value, mapping to new `value` property.

Note: this could be relevant if some tenat had previous third party integration that use this specific event, and this specific value.

talbertosilva: Fixed! I've added a verification on `getEventTotalValue` method to check if total exists and, if so, map that to `value` property.

